PR #118, opened 2022-04-01
change unit according to the target device

PikachuHy:


PikachuHy:
Tested only on macos

PikachuHy:
the `retina_factor` variable only works on macOS
in other platforms, its value is 1.

src/Plugins/Qt/qt_utilities.cpp
```diff
@@ -427,6 +427,11 @@ qt_native_image_size (url image, int& w, int& h) {
 
 void
 qt_pretty_image_size (int ww, int hh, string& w, string& h) {
+  if (get_preference("target device") == "screen") {
```
darcy-shen, line 430: could you describe when the "target_device" is screen

darcy-shen: it seems it is related to macOS

darcy-shen: Here is how we tuning for macOS
```
#ifdef OS_MACOS
code that only work for macOS
#endif
```

PikachuHy: > when the "target_device" is screen

In my mind
If you are confused about the target device then it should be the screen.
If you prefer to use px, your target device should be the screen.


Generally, the default unit `pt` is used in print typefaces, while the unit `px` is used on the screen.

see [Should I use pt or px](https://stackoverflow.com/questions/3557260/should-i-use-pt-or-px) for more detail.

PikachuHy: > why you need to make these changes

When pasting an image from the clipboard, the unit of image size is `px`.
However, the default behavior of Mogan(TeXmacs) is to change the unit to `pt`, which causes the image scaled and blurred.

Blurry images are always annoying.

For most of us, we should keep the image size in Mogan the same as in the clipboard.

PikachuHy: > it seems it is related to macOS

No, it is related to Qt.

This change only affects Qt Interface.

image

The `retina_factor` variable only works on macOS.
Most of the time, its value is 2.
In other operating systems, its default value is 1.
So, the code has no side effects on Windows or Linux.

> tuning for macOS

We also don't need to add a guard.

src/Plugins/Qt/qt_utilities.cpp
```diff
@@ -427,15 +427,19 @@ qt_native_image_size (url image, int& w, int& h) {
 
 void
 qt_pretty_image_size (int ww, int hh, string& w, string& h) {
-  SI pt = get_current_editor()->as_length ("1pt");
+  string unit = "pt";
```
darcy-shen, line 430: we'd better follow the jolly style:
+ assignment: `a= 1` but not `a = 1`
+ func apply: `func ()` but not `func()`

darcy-shen: not all code in the TeXmacs code base are in jolly style. But we'd better follow the jolly style.

darcy-shen: Because it's Joris who created the AWESOME GNU TeXmacs.

darcy-shen (approved):
LGTM. The only difference is in macOS for retina screen.

With the default target device, the behavior is the same.

The only thing you need to do is to adapt your code into jolly style.

src/Plugins/Qt/qt_utilities.cpp
PikachuHy: It would be nice if we could write a format config file like `.clang-format`.

darcy-shen: absolutely true, but sometimes I think it wastes too much time to just make ci happy as a human being

just remember, it is better to write in jolly style, but it is not a strict rule

If we make a clang-format rule, we should bring it to the community. And format the existing coding base at once.

And then we need to rename current main branch to mogan1, and start from the scratch with the latest GNU TeXmacs release.

PR #119, opened 2022-04-01
Use libcurl instead for invoke the curl command line

darcy-shen:
## How to test it
Launch mogan via terminal, check on `Debug -> io`.

And then try to open a web file in `Help -> Mogan`. The following debugging meesage will be printed in the console.

```
TeXmacs] debug-io, curl --user-agent TeXmacs-2.1.1 https://gitee.com/XmacsLabs/mogan/raw/main/TeXmacs/doc/about/xmacs/2019年全国高中数学联合竞赛一试试题A卷.tm --output /home/sadhen/.Xmacs/system/tmp/93518/tmp_1951389520
```

## Why
The current impl actually work only on macOS and Linux. In this pull request, I decide only replace the curl command line invoke with libcurl. And later, I will submit the follow-up pull requests to:
+ make it work on Windows
+ correctly handle the exceptions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,6 +273,7 @@ find_package (PNG)
 find_package (Iconv)
 find_package (ZLIB)
 find_package (JPEG)
+find_package (CURL)
 #FIND_PACKAGE(Axel)
 
 # Allow for selection of Scheme implementation
@@ -396,6 +397,7 @@ endif (WIN32)
 set (TeXmacs_Include_Dirs ${TeXmacs_Include_Dirs}
   ${FREETYPE_INCLUDE_DIRS} ${Cairo_INCLUDE_DIRS}
   ${IMLIB2_INCLUDE_DIR} ${GMP_INCLUDES}
+  ${CURL_INCLUDE_DIRS}
 )
 
 ### --------------------------------------------------------------------
@@ -434,6 +436,7 @@ file (GLOB_RECURSE TeXmacs_Std_Plugins_SRCS
   "${TEXMACS_SOURCE_DIR}/src/Plugins/Openssl/*.cpp"
   "${TEXMACS_SOURCE_DIR}/src/Plugins/Sqlite3/*.cpp"
   "${TEXMACS_SOURCE_DIR}/src/Plugins/Updater/*.cpp"
+  "${TEXMACS_SOURCE_DIR}/src/Plugins/Curl/*.cpp"
 )
 
 file (GLOB_RECURSE TeXmacs_Style_SRCS
@@ -494,6 +497,7 @@ set (TeXmacs_Libraries
   ${ZLIB_LIBRARIES}
   ${JPEG_LIBRARIES}
   ${PNG_LIBRARIES}
+  ${CURL_LIBRARIES}
   -lpthread
 )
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -273,6 +273,7 @@ find_package (PNG)
 find_package (Iconv)
 find_package (ZLIB)
 find_package (JPEG)
+find_package (CURL)
```
PikachuHy, line 276: use target instead of global variables `CURL_INCLUDE_DIRS`, `CURL_LIBRARIES`

```cmake
target_link_library(your-target PUBLIC CURL::libcurl)
```

see https://cmake.org/cmake/help/latest/module/FindCURL.html?highlight=curl#imported-targets

PikachuHy: if `curl` is a necessary dependency, add arg `REQUIRED` to `find_package`.

```
find_package(CURL REQUIRED)
```

PikachuHy: what's your CMake version in the development machine?

src/Plugins/Curl/curl.cpp
```diff
@@ -9,3 +9,28 @@
...
+static size_t write_data(void *ptr, size_t size, size_t nmemb, FILE *stream) {
+    size_t written = fwrite(ptr, size, nmemb, stream);
+    return written;
+}
+
+void curl_download(string source, string target, string user_agent) {
+  debug_io << "curl --user-agent " << user_agent << " "
+           << source << " --output " << target << LF;
+  CURL *curl= curl_easy_init ();
+  if (curl) {
```
PikachuHy, line 24: what if curl init fails or download file fails?

src/System/Files/web_files.cpp
```diff
@@ -70,41 +71,21 @@ web_encode (string s) {
...
-    cmd << " " << escape_sh (web_encode (as_string (name)));
-    cmd << " --output " << tmp_s;
-  }
+  curl_download(
+    escape_sh (web_encode (as_string (name))),
+    tmp_s,
+    string("TeXmacs-") * TEXMACS_VERSION);
 
   //cout << cmd << LF;
   system (cmd);
```
PikachuHy, line 91: remove unnecessary code

the cmd is empty.

src/System/Files/web_files.cpp
```diff
@@ -70,41 +71,21 @@ web_encode (string s) {
...
   if (!is_rooted_web (name)) return url_none ();
   url res= get_cache (name);
   if (!is_none (res)) return res;
 
-  string tool= fetch_tool ();
-  if (tool == "") return url_none ();
-  
   url tmp= url_temp ();
   string tmp_s= escape_sh (concretize (tmp));
   string cmd= "";
```
PikachuHy, line 82: delete line

PikachuHy (changes requested):

CMakeLists.txt
darcy-shen:
```
$ cmake --version
cmake version 3.23.0
```

darcy-shen: For the `target_link_library` syntax, we can adapt all the occurrences in the follow-up pr.

src/Plugins/Curl/curl.cpp
darcy-shen: now, I will print out the error message below

PikachuHy (approved):
LGTM

PikachuHy:
image

please use squash to eliminate redundant commits before merge

and attach the pr url to commit message

PR #121, opened 2022-04-02
copy image to current folder when pasting image from the clipboard

PikachuHy:
## Overview

Give an option like Typora,

image

But now, only support `copy image to current folder`.

## How to test
add the following scheme code to your `init-buffer.scm`

```scm
(define (notify-julia-syntax2 var val)
  ())

(define-preferences
  ("copy image to current folder" "on" notify-julia-syntax2))
```
Is there a more elegant way to change user preferences? 😭

## Why

I just want to provide an option
separate binary data and plain text.

## snapshot

image

### DIFF
--- a/src/Plugins/Qt/qt_gui.cpp
+++ b/src/Plugins/Qt/qt_gui.cpp
@@ -305,7 +305,7 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
   }
   if (input_format == "picture") {
     tree t (IMAGE);
-    auto img = qvariant_cast<QImage>(md->imageData());
+    QImage img = qvariant_cast<QImage>(md->imageData());
     QSize size= img.size ();
     int ww= size.width (), hh= size.height ();
 #ifdef OS_MACOS
--- a/src/Plugins/Qt/qt_gui.cpp
+++ b/src/Plugins/Qt/qt_gui.cpp
@@ -47,6 +47,8 @@
 #include <QUrl>
 #include <QDesktopWidget>
 #include <QApplication>
+#include <QDateTime>
+#include <QFileInfo>
 
 #include "QTMGuiHelper.hpp"
 #include "QTMWidget.hpp"
@@ -303,7 +305,8 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
   }
   if (input_format == "picture") {
     tree t (IMAGE);
-    QSize size= qvariant_cast<QImage>(md->imageData()).size ();
+    auto img = qvariant_cast<QImage>(md->imageData());
+    QSize size= img.size ();
     int ww= size.width (), hh= size.height ();
 #ifdef OS_MACOS
     ww /= retina_factor;
@@ -311,7 +314,23 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
 #endif
     string w, h;
     qt_pretty_image_size (ww, hh, w, h);
-    t << tuple (tree (RAW_DATA, s), "png") << w << h << "" << "";
+    if (get_preference("copy image to current folder") == "on") {
+      string doc_path = as_string(get_current_buffer());
+      // e.g. 20220402114852693Sat
+      QString cur_time_str = QDateTime::currentDateTime().toString("yyyyMMddhhmmsszzzddd");
+      QString tmp_image_name = QString("img_%1.png").arg(cur_time_str);
+      QString tmp_image_dir = QFileInfo(to_qstring(doc_path)).absolutePath();
+      QString tmp_image_path = QString("%1/%2").arg(tmp_image_dir).arg(tmp_image_name);
+      bool ok = img.save(tmp_image_path);
+      if (ok) {
+        t = tree(IMAGE, from_qstring(tmp_image_name), w, h, "", "");
+      } else {
+        // insert raw data if fail
+        t << tuple (tree (RAW_DATA, s), "png") << w << h << "" << "";
+      }
+    } else {
+      t << tuple (tree (RAW_DATA, s), "png") << w << h << "" << "";
+    }
     s= as_string (call ("convert", t, "texmacs-tree", "texmacs-snippet"));
   }
   if (input_format == "linked-picture") {
--- a/src/Plugins/Qt/qt_gui.cpp
+++ b/src/Plugins/Qt/qt_gui.cpp
@@ -316,8 +316,8 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
     qt_pretty_image_size (ww, hh, w, h);
     if (get_preference("copy image to current folder") == "on") {
       string doc_path = as_string(get_current_buffer());
-      // e.g. 20220402114852693Sat
-      QString cur_time_str = QDateTime::currentDateTime().toString("yyyyMMddhhmmsszzzddd");
+      // e.g. 20220402114852693
+      QString cur_time_str = QDateTime::currentDateTime().toString("yyyyMMddhhmmsszzz");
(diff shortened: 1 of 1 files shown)


darcy-shen:
#118 also needs a option in the preference. Well, I think it is ok to impl the feature first and then add the preferences.

src/Plugins/Qt/qt_gui.cpp
```diff
@@ -303,15 +305,32 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
   }
   if (input_format == "picture") {
     tree t (IMAGE);
-    QSize size= qvariant_cast<QImage>(md->imageData()).size ();
+    auto img = qvariant_cast<QImage>(md->imageData());
```
darcy-shen, line 308: it is ok to use C++11 syntax in Mogan.

PikachuHy:
?

darcy-shen:
I have to test it locally on my laptop. Yesterday, I just reviewed the code.

src/Plugins/Qt/qt_gui.cpp
```diff
@@ -303,15 +305,32 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
...
     ww /= retina_factor;
     hh /= retina_factor;
 #endif
     string w, h;
     qt_pretty_image_size (ww, hh, w, h);
-    t << tuple (tree (RAW_DATA, s), "png") << w << h << "" << "";
+    if (get_preference("copy image to current folder") == "on") {
+      string doc_path = as_string(get_current_buffer());
+      // e.g. 20220402114852693Sat
+      QString cur_time_str = QDateTime::currentDateTime().toString("yyyyMMddhhmmsszzzddd");
```
darcy-shen, line 320: image

I'm using the Chinese locale, is it the expected file name?

src/Plugins/Qt/qt_gui.cpp
```diff
@@ -303,15 +305,32 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
   }
   if (input_format == "picture") {
     tree t (IMAGE);
-    QSize size= qvariant_cast<QImage>(md->imageData()).size ();
+    QImage img = qvariant_cast<QImage>(md->imageData());
```
darcy-shen, line 308: please use jolly style

src/Plugins/Qt/qt_gui.cpp
```diff
@@ -303,15 +305,32 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
...
     int ww= size.width (), hh= size.height ();
 #ifdef OS_MACOS
     ww /= retina_factor;
     hh /= retina_factor;
 #endif
     string w, h;
     qt_pretty_image_size (ww, hh, w, h);
-    t << tuple (tree (RAW_DATA, s), "png") << w << h << "" << "";
+    if (get_preference("copy image to current folder") == "on") {
+      string doc_path = as_string(get_current_buffer());
```
darcy-shen, line 318: if the current buffer is the no name file, we will save the image to `$TEXMACS_HOME_PATH/texts/scratch`

src/Plugins/Qt/qt_gui.cpp
```diff
@@ -303,15 +305,32 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
...
+      // e.g. 20220402114852693Sat
+      QString cur_time_str = QDateTime::currentDateTime().toString("yyyyMMddhhmmsszzzddd");
+      QString tmp_image_name = QString("img_%1.png").arg(cur_time_str);
+      QString tmp_image_dir = QFileInfo(to_qstring(doc_path)).absolutePath();
+      QString tmp_image_path = QString("%1/%2").arg(tmp_image_dir).arg(tmp_image_name);
+      bool ok = img.save(tmp_image_path);
+      if (ok) {
+        t = tree(IMAGE, from_qstring(tmp_image_name), w, h, "", "");
+      } else {
+        // insert raw data if fail
```
darcy-shen, line 328: Use debug_io to print out the `tmp_image_path` here?

It should happen really rarely.

src/Plugins/Qt/qt_gui.cpp
darcy-shen: If the current buffer is the immutable doc file, I think we will fail to save it to the corresponding folder.

darcy-shen: image

And sometimes, the current buffer is actually a tmfs path.

darcy-shen: For remote files, the current files are using http/https protocal:

image

darcy-shen (changes requested):
If a pr is not ready to be reviewed, just mark it as draft. I've just marked your cmake pr as draft.

src/Plugins/Qt/qt_gui.cpp
PikachuHy: no, I will remove `ddd`

src/Plugins/Qt/qt_gui.cpp
PikachuHy: > the current buffer is actually a tmfs path.

```
bool ok= img.save (tmp_image_path);
```
`img.save` will return false.

So, we will insert raw data

PikachuHy: > If the current buffer is the immutable doc file (doc files under $TEXMACS_HOME), I think we will fail to save it to the corresponding folder.

How to test that?

if saving images fail, we will insert raw data.

darcy-shen: just test if a URL is starting with $TEXMACS_HOME

darcy-shen: See #132 , if you are not familiar with the API, add unit tests for them first.

src/Plugins/Qt/qt_gui.cpp
```diff
@@ -303,15 +305,42 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
...
     ww /= retina_factor;
     hh /= retina_factor;
 #endif
     string w, h;
     qt_pretty_image_size (ww, hh, w, h);
-    t << tuple (tree (RAW_DATA, s), "png") << w << h << "" << "";
+    bool need_save_img= (get_preference("copy image to current folder") == "on");
+    string doc_path= as_string (get_current_buffer ());
+    // only local document save image
+    if (starts (doc_path, "tmfs://")
```
darcy-shen, line 320:
``` cpp
need_save_image = (pref == "on") && !is_rooted_web(url) && !is_rooted_tmfs(url) && descends(url, url_system("$TEXMACS_PATH"))
```

src/Plugins/Qt/qt_gui.cpp
```diff
@@ -303,15 +305,39 @@ qt_gui_rep::get_selection (string key, tree& t, string& s, string format) {
...
 #endif
     string w, h;
     qt_pretty_image_size (ww, hh, w, h);
-    t << tuple (tree (RAW_DATA, s), "png") << w << h << "" << "";
+    url doc_url= get_current_buffer ();
+    // only local document save image
+    if ((get_preference ("copy image to current folder") == "on")
+        && !is_rooted_web (doc_url)
+        && !is_rooted_tmfs (doc_url)
+        && descends (doc_url, url_system ("$TEXMACS_PATH"))) {
```
darcy-shen, line 322: !descends

doc_url should not start with $TEXMACS_PATH

darcy-shen (approved):
manually tested on GNU/Linux

Issue #122, opened 2022-04-02
CMake: use target instead of global variables


darcy-shen:
use target instead of global variables CURL_INCLUDE_DIRS, CURL_LIBRARIES

```
target_link_library(your-target PUBLIC CURL::libcurl)
```
as discussed in #119 .

PikachuHy:
I will rewrite the `CMakeLists.txt` with a modern cmake style.

require at least cmake 5.15

darcy-shen:
xmake seems to be a better alternative of CMake/Autotools.

We may switch to xmake when we start to work on Mogan 2.x.

https://xmake.io/#/

darcy-shen:
closed by #131

PR #123, opened 2022-04-03
Rewrite cmake

PikachuHy:
see https://github.com/XmacsLabs/mogan/issues/122

DONE

- upgrade the minimum CMake version to 5.15
- set default c++ standard to 17, although TeXmacs don't use c++17 features.
- for all dependencies, use the target-based style, which can manage headers and libraries automatically.
- initial support for Qt6. However, some APIs in Qt6 have changed. We need some guards or rewrite some code.
- delete old `FindXXX.cmake`. The latest CMake has already contained them.
- two extra empty `.cpp` files. Usually, one Qt class, one `.h/.cpp` file. 

TODO

- rewrite `FindCairo.cmake`
- rewrite `FindGMP.cmake`
- rewrite `FindImlib2.cmake`

It seems that the Mogan does not use any of them.
The GMP is not a necessary dependency for S7.

- eliminate hardcode config, e.g. `CONFIG_HOST_OS`
- eliminate redundant prefix `${TEXMACS_SOURCE_DIR}` when collecting source code files.

The file path will be handled by CMake automatically after 5.15.


darcy-shen:
I suggest that you should break down this pull requests into small PRs. Small PRs are more friendly for reviewers.

darcy-shen:
I suggest that you should break down this pull requests into small PRs. Small PRs are more friendly for reviewers.

PikachuHy:
first, replace all global variables with targets

https://github.com/XmacsLabs/mogan/pull/131

PikachuHy:
second, upgrade CMake version
https://github.com/XmacsLabs/mogan/pull/134

PikachuHy:
third, upgrade c++ standard
https://github.com/XmacsLabs/mogan/pull/137

PR #130, opened 2022-04-04
Fix: do not use the cat command to check if it is empty

darcy-shen:


PikachuHy (approved):
LGTM

please use 
image

darcy-shen:
I just merged #132 using `Rebase and merge`. It messed up the main branch with 4 commits.

I think we should use `Squash and merge`.

PR #134, opened 2022-04-05
upgrade cmake version

PikachuHy:


PikachuHy:
sub pr of https://github.com/XmacsLabs/mogan/pull/123

darcy-shen:
Why we need to upgrade the CMake version? What feature does CMake 3.15 brings?

PikachuHy:
> Why we need to upgrade the CMake version? What feature does CMake 3.15 brings?

after CMake 3.15, the target-based feature is considered relatively complete.

darcy-shen (approved):
LGTM

PR #136, opened 2022-04-05
initial support for Qt 6

PikachuHy:
Build TeXmacs with Qt6

maybe some bugs, we need more tests.

All changes do not affect the Qt5 version.

## How to test
set `CMAKE_PREFIX_PATH` to Qt6 cmake path

```
mkdir build
cd build
cmake -DCMAKE_PREFIX_PATH=/Users/pikachu/Qt/6.3.0/macos/lib/cmake ..
make -j4
```

## TODO

There are no equivalent replacements for the following changes.

- QPrinter::setPaperSize

no `QPrinter::DevicePixel` in QPageSize

on TeXmacs qt6 branch, it just print log
```
if (DEBUG_CONVERT) debug_convert << "no dpi or size informations!!" << LF;
```

- QWheelEvent::delta

- QApplication::globalStrut

on TeXmacs qt6 branch, a fixed value assigned
```
calcSize = QSize(2, 2);
```

## snapshot
After merge https://github.com/XmacsLabs/mogan/pull/135

We can see

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -546,7 +546,7 @@ if (TEXMACS_GUI MATCHES "Qt.*")
             Qt${QT_VERSION_MAJOR}::PrintSupport
             Qt${QT_VERSION_MAJOR}::Svg
             )
-    if (APPLE)
+    if (APPLE AND ${QT_VERSION_MAJOR} LESS_EQUAL 5)
       find_package (Qt5 COMPONENTS MacExtras REQUIRED)
       set (QT_LIBRARIES ${QT_LIBRARIES} Qt5::MacExtras)
     endif ()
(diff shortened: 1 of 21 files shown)


PikachuHy:
Qt6 in TeXmacs

https://lists.gnu.org/archive/html/texmacs-dev/2021-06/msg00096.html

https://github.com/texmacs/texmacs/commits/qt6

darcy-shen:
What's the main difference between Qt 5 and Qt 6 for GNU TeXmacs?

darcy-shen:
Personally, I do not want to developing on two Qt versions. I suggest that we should use only one of the Qt version.

How about switching to Qt 6? Is there any blocking issues?

darcy-shen:
https://doc.qt.io/qt-6/supported-platforms.html

No 32bit support.

PR #141, opened 2022-04-06
add .clang-format

PikachuHy:
clang format config 

> we'd better follow the jolly style:
assignment: a= 1 but not a = 1
func apply: func () but not func()

https://github.com/XmacsLabs/mogan/pull/118#discussion_r841077011

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,2 +1,11 @@
 SpaceBeforeAssignmentOperators: false
 SpaceBeforeParens: Always
+AlwaysBreakAfterReturnType: TopLevel
+AllowShortIfStatementsOnASingleLine: AllIfsAndElse
+AlignConsecutiveAssignments: Consecutive
+AlignConsecutiveDeclarations: Consecutive
+SpaceAfterCStyleCast: true
+AllowShortFunctionsOnASingleLine: All
+BreakBeforeBraces: Custom
+BraceWrapping:
+  BeforeElse: true
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,2 @@
+SpaceBeforeAssignmentOperators: false
+SpaceBeforeParens: Always
--- a/src/Texmacs/Texmacs/texmacs.cpp
+++ b/src/Texmacs/Texmacs/texmacs.cpp
@@ -1,33 +1,34 @@
 
 /******************************************************************************
-* MODULE     : texmacs.cpp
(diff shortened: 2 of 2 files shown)


PikachuHy:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

darcy-shen:
https://github.com/XmacsLabs/mogan/blob/main/src/Texmacs/Texmacs/texmacs.cpp#L91-L95
```
void 
clean_exit_on_segfault (int sig_num) {
  (void) sig_num;
  FAILED ("segmentation fault");
}
```
The return type should be put on the first line. And the second line followed by the function name.

darcy-shen:
I suggest that you should add a script named `format` under `bin/`, in that way:

```
bin/format src/Texmacs/Texmacs/texmacs.cpp
# <=> clang-format -i src/Texmacs/Texmacs/texmacs.cpp
```
Please also commit the result of `bin/format src/Texmacs/Texmacs/texmacs.cpp`, and try to reduce unnecessary changes as much as possible.

Once it is ok, it would be better to start a discussion on the texmacs-dev mail list.

.clang-format
```diff
@@ -0,0 +1,2 @@
+SpaceBeforeAssignmentOperators: false
+SpaceBeforeParens: Always
```
darcy-shen, line 2:
```
AlwaysBreakAfterReturnType: TopLevelDefinitions
```
Since  clang-format 3.8

darcy-shen:
```
AllowShortIfStatementsOnASingleLine: AllIfsAndElse
```
Since clang-format 9

darcy-shen:
```
AlignConsecutiveAssignments: true
```

darcy-shen:
```
SpaceAfterCStyleCast: true
```

PikachuHy: > AllowShortIfStatementsOnASingleLine: AllIfsAndElse
> AllowShortFunctionsOnASingleLine: All

annoying rule for debugging

darcy-shen: > annoying rule for debugging

I think so! Personally, I prefer the verbose one.

darcy-shen: If there is no explicit good reason, I would like to respect the history code and follow the convention.

darcy-shen (approved):
LGTM

Issue #4, opened 2021-12-19
UI needs to be updated


YiqiXu:
The user interface of TeXmacs doesn't change much over the last two decades. According to my personal experiences, the outdated UI did constantly push away new users interested in TeXmacs. I think that modernizing the UI should be a long-term goal of Mogan.

The following two issues could be solved in a short term.

1. Updating the icons on the menu seems a viable option and a great improvement in the short term.
2. The dark mode is far beyond ideal for daily usage, for example, the following screenshot of my MacBook air M1.
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/32867606/146690686-e3cdecb6-a22e-4618-9710-5999d802f7c0.png">

darcy-shen:
We could use the icons from the KDE project.

darcy-shen:
derived issues: #26 and #27

fbob:
Speaking about the UI, some other improvements could be:
- a Command Palette to access all the features of Texmacs without having to remember key strokes.
- a TOC panel to be able to switch to a particular heading and move them easily. Can be useful for slides too.
- inline text, math symbols completion with dropdown list
- inline spellchecking correction propositions with a dropdown list
- tabs like firefox tabs for document management
-  split views to see more documents at the same time or different parts of the same document
- ....

darcy-shen:
I think the dropdown list for completion is the most urgent one.

PikachuHy:
document tab is here. https://github.com/XmacsLabs/mogan/pull/142

fbob:
Nice one ! :+1:

da-liii:
Since Mogan STEM v2025.1.0, there will be a modern UI available.

And in Liii STEM v2025.0.11, the modern UI called `liii` is available now!

Issue #60, opened 2022-01-09
It fails to open all files in one window as tabs.


YiqiXu:
TeXmacs v2.1.1 can open all files in one window as tabs as in the following screenshot, only when both the following options are turned on: 

- `always` under `System Preferences -> General -> Prefer tabs` in the app Preview on macOS, on the one hand, and
-  `Documents in separate windows` under `Edit -> Preferences -> General -> Buffer managements` in TeXmacs, on the other hand. 

But in Mogan v1.0 alpha 4, it fails.

<img width="704" alt="image" src="https://user-images.githubusercontent.com/32867606/148685276-0a91758c-3b14-474f-8fda-7697bea0aa56.png">

medwatt:
@YiqiXu, how did you manage to display opened buffers as tabs? I'm on Linux and don't seem to have a `Preview` menu.

YiqiXu:
By `Preview` I mean the internal app Preview on macOS. I have coincidentally found out that we could get what we want under such tricky settings. As I only use TeXmacs on Windows and macOS, I have no idea how to reproduce it on Linux. Sorry for being unclear, and I have changed the description accordingly.

YiqiXu:
This issue is resolved in #142.

darcy-shen:
@PikachuHy it seems it works only for macOS. We still need to implement the QTMDocTab for GNU/Linux.

darcy-shen:
Let us move QTMDocTab above the mode tool bar first and then fix bugs for QTMDocTab.

PikachuHy:
ok.
disable `QTMDocTab` on macOS first.

YiqiXu:
Update: the setting in the description is no longer working in TeXmacs 2.1.1 M1 version. 
@PikachuHy @darcy-shen, I think we'd better reopen #142.

PR #142, opened 2022-04-06
support Safari-like tab

PikachuHy:
https://github.com/XmacsLabs/mogan/issues/4

> tabs like firefox tabs for document management

## FIXME

- document changes are not smooth

although using the same API as `go-menu`

```
call ("switch-to-buffer*", object (m_buffers[index]));
```

## snapshot

- Safari

image

- TeXmacs

image

### DIFF
--- a/src/Plugins/Qt/DocTabBar.h
+++ b/src/Plugins/Qt/DocTabBar.h
@@ -1,6 +1,12 @@
-//
-// Created by PikachuHy on 2022/4/6.
-//
+/******************************************************************************
+* MODULE     : texmacs.cpp
+* DESCRIPTION: Document Tab Bar
+* COPYRIGHT  : (C) 2022  PikachuHy
+*******************************************************************************
+* This software falls under the GNU general public license version 3 or later.
+* It comes WITHOUT ANY WARRANTY WHATSOEVER. For details, see the file LICENSE
+* in the root directory or <http://www.gnu.org/licenses/gpl-3.0.html>.
+******************************************************************************/
 
 #ifndef TEXMACS_DOCTABBAR_H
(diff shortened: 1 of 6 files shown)


src/Plugins/Qt/DocTabBar.h
```diff
@@ -0,0 +1,24 @@
+//
+// Created by PikachuHy on 2022/4/6.
```
darcy-shen, line 2: same above

src/Plugins/Qt/DocTabBar.h
```diff
@@ -0,0 +1,24 @@
...
+//
+
+#ifndef TEXMACS_DOCTABBAR_H
+#define TEXMACS_DOCTABBAR_H
+#include "url.hpp"
+#include <QList>
+#include <QTabBar>
+class DocTabBar : public QTabBar {
```
darcy-shen, line 10: The name convention should be QTMDocTabBar

YiqiXu:
Once this PR is merged, issue #60 can be closed.

YiqiXu:
Is it controlled by the following option?

image

darcy-shen:
> Do the following options control it? Or should we add another option like "Multiple documents in tabs"?

Let's merge the existing code first and then we can improve it step by step. Just cannot wait to use it!

darcy-shen (changes requested):
Please follow the class name convention of the original GNU TeXmacs. The reason is that we'd better contribute this pull request back to GNU TeXmacs.

PikachuHy:
warning

this feature may cause the Qt6 version TeXmacs to throw an exception.

weird

darcy-shen (approved):
LGTM

PR #147, opened 2022-04-07
[no ci] gnuplot: set default width to 0.8par

darcy-shen:
> Tips: use `[no ci]` to mute github action when we actually do not need it.

See https://github.com/XmacsLabs/mogan/pull/141/commits, the first commit should use `[no ci]`.

> Why 0.8par?

Because `0.8par` is a good width for session, foldable session and big picture. Setting `Document -> Paragraph... -> Number of columns` to 2, in this case, `1par` is a better one. We can take advantage of dynamic default width later when we begin to make a significant change of those graphics plugins.


plugins/tmpy/graph/graph.py
```diff
@@ -22,10 +22,12 @@ class Graph(object):
     message = ""
     pre_code = ""
     post_code = ""
-    height = 0
-    width = 0
-    default_output = "eps"
+    height = ""
+    width = ""
     output = ""
+    default_width = "0px"
```
darcy-shen, line 28: For compatible reason, this pr is only for gnuplot.

PikachuHy (approved):
LGTM

PR #150, opened 2022-04-08
[no ci] Gnuplot: refactor and disable buggy input converter

### LINKED ISSUE #148: gnuplot: failed to use the magic line
```
% -width 1par
plot sin(x)
```

Here is the error message:
```
parplot is not allowed, please pt, px or par!
```

darcy-shen:
closes #148 

+ Do not use `~` as the separator
+ Disable buggy input converter (I guess that's why `~` is needed, the current input converter swallows newline)
+ Re-use the main-loop


PikachuHy (approved):
LGTM

PR #158, opened 2022-04-16
CMake: set CONFIG_HOST_OS

darcy-shen:
Tested via `Debug -> Misc -> Provoke C++ Error`


PikachuHy (approved):
LGTM

how about `CMAKE_SYSTEM_NAME `?

https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_NAME.html#variable:CMAKE_HOST_SYSTEM_NAME

PR #160, opened 2022-04-17
Deprecate Traditional Chinese terms

darcy-shen:
## Why

+ Traditional Chinese is not required for Science and Technology.
+ I dislike the more Japanese-like terms adopted in Taiwan and Traditional Chinese is not in Taiwan.

## What
1. Removing `TeXmacs/examples/texts/torture-images.tm`, because I failed to open it
2. Deprecate `chineset`

## TODO
Mogan should not crash when opening ill-formatted tm files, like unsupported languages.


YiqiXu (commented):
I'm afraid I have to disagree that traditional Chinese is not required for science and technology. I know some mathematicians from Hongkong and Taiwan use traditional Chinese. I think Mogan should be as inclusive as possible so that we can attract more users, not the other way around. 

I still suggest changing `Taiwanese` to `Traditional Chinese`.

By the way, I don't know what you mean by Japanese-like terms adopted in Taiwan.

darcy-shen:
Sorry, I was biased.

PR #179, opened 2022-04-27
Re-impl read_directory using Qt

darcy-shen:
IO code in https://github.com/XmacsLabs/mogan/blob/main/src/System/Files/file.cpp is not in a good quality. We should replace it using Qt.

Without it crashes on Windows because `read_directory` on Windows returns messy code.

### DIFF
--- a/src/Plugins/Qt/qt_file.cpp
+++ b/src/Plugins/Qt/qt_file.cpp
@@ -1,0 +1,35 @@
+
+/******************************************************************************
+* MODULE     : qt_file.cpp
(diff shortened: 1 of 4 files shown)


PikachuHy:
What's the difference between filename `qt_file.cpp` and `QTMFile.cpp`?

PikachuHy (approved):
LGTM

darcy-shen:
`qt_file.cpp` means it is a Qt impl of `file.cpp`. `QTMxxx` is Qt code, and is not related to the TeXmacs abstraction.

PR #187, opened 2022-04-29
[no ci] S7: LaTeX import fix

darcy-shen:


darcy-shen:
@mgubi I just commited your S7 fix on LaTeX export, see #194 

Your fix is better than my fix via #192 

And this is my fix for LaTeX import. It is not a very clean fix, but it works.

``` scheme
(string->symbol "")
```
will fail in S7. So, I fixed it by using `(string->symbol " ")`

darcy-shen:
I think GNU TeXmacs with S7 is good enough to be adopted in the official repo soon.

I will release Mogan v1.0.2  (migrated to S7) on 2022/05/04. Let me gather some user feedbacks. S7 is awesome!

mgubi:
Just a (small?) warning: S7 in TeXmacs require a small patch which according to the main developer of S7 is not totally correct. I haven't experienced any problem with it, but maybe we should be a bit cautious. Unfortunately without the patch the symbol lookup in S7 become too slow for reasons which are not clear neither to the main developer nor to me. We could help him to debug the issue and hope we can find a better fix.

PR #232, opened 2022-05-18
Fix slot_name

darcy-shen:


src/Edit/Interface/edit_interface.hpp
```diff
@@ -11,6 +11,7 @@
 
 #ifndef EDIT_INTERFACE_H
 #define EDIT_INTERFACE_H
+
```
darcy-shen, line 14: just want to make it better

src/Graphics/Gui/widget.cpp
```diff
@@ -39,6 +39,7 @@ slot_name (const slot s) {
     "SLOT_MOUSE_POINTER",
     "SLOT_INVALIDATE",
     "SLOT_INVALIDATE_ALL",
+    "SLOT_INVALID",
```
darcy-shen, line 42: https://github.com/XmacsLabs/mogan/blob/6c6bf6be2fe912de44c78c2ed189255df87cd705/src/Graphics/Gui/message.hpp#L43:L42

PikachuHy: how `SLOT_INVALID` work?

darcy-shen: I know about the slots, but I have not read the code about every slot.

It is a design of the server/client arch.

PikachuHy (approved):
LGTM

PR #245, opened 2022-05-24
Qt: fix QTMLineEdit crash when using Input Method

### LINKED ISSUE #222: Searching for Chinese is not working correctly.
In Mogan v.1.0.2, searching for Chinese is not working correctly.

darcy-shen:
Fixes #222 

It fixes the crash issue, and we still need to click twice to open the document searching window.


PikachuHy (approved):
LGTM

PR #246, opened 2022-05-24
[no ci] Open links in remote files as in local files

### LINKED ISSUE #234: Links in remote TeXmacs files should be opened properly
Just like links in local TeXmacs file:
+ Open link ending with tm as TeXmacs file
+ Using external browsers to open other links

darcy-shen:
Links open rules:
1. If it is a http/https/ftp/... link
   + if it ends with tm, open it in Mogan Editor
   + else just open it in web browser
2. If it is a local file, open it in Mogan Editor

Before, the rules work fine when we are in a local file buffer but it failed to work in a remote file buffer.
With the PR merged, the rules is the same for both local and remote files.

Fixes #234


PikachuHy (approved):
LGTM

PR #267, opened 2022-05-30
Using got_focus instead of the bitwise operator

darcy-shen:
git grep `got_focus=`, we can find all assignment of got_focus


PikachuHy (approved):
LGTM

PR #282, opened 2022-06-14
A workaround to attach the style 'code' automatically

darcy-shen:


darcy-shen:
I'm confident on the second attempt to attach the style "code".

PR #290, opened 2022-06-25
CMake: for debian and macOS development

darcy-shen:
rework of #243


darcy-shen:
Verified on macOS and Linux, both work fine for dev purpose

PR #297, opened 2022-07-10
option for TM format with UTF8 support

darcy-shen:
The reader is compatible with the previous reader.

For the writer, we provide the option `tm format with utf8`, it will be default to off for GNU TeXmacs.

### DIFF
--- a/src/Data/Convert/Texmacs/totm.cpp
+++ b/src/Data/Convert/Texmacs/totm.cpp
@@ -144,6 +144,12 @@ tm_writer::write (string s, bool flag, bool encode_space) {
     for (i=0; i<n; i++) {
       char c= s[i];
       if ((c == ' ') && (!encode_space)) write_space ();
+      else if (c == '<' && i+6 < n && s[i+1] == '#' && s[i+6] == '>') {
+        tmp << cork_to_utf8 (s(i, i+6));
+        i+= 6;
+        spc_flag= false;
+        ret_flag= false;
+      }
       else {
         if (c == ' ') tmp << "\\ ";
         else if (c == '\n') tmp << "\\n";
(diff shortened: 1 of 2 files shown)


darcy-shen:
https://github.com/sheredom/utf8.h

@PikachuHy this header-only impl seems to be a good impl.

darcy-shen:
another impl: https://github.com/JuliaStrings/utf8proc

PikachuHy:
a simple way to handle unicode

light-weight and useful

https://github.com/NanoMichael/MicroTeX/blob/da9ed490c5f21a10914a3b222099b7a3b102478f/lib/utils/utf.cpp#L94


```c++

sptr<Atom> Parser::getCharAtom() {
  int n = 0, m = 0, cnt = 0;
  c32 chr = 0;
  const auto next = [&]() {
    return microtex::nextUnicode(_latex, _pos + n, m);
  };
  const auto collect = [&](c32 code) {
    n += m;
    cnt++;
    chr = code;
  };
  microtex::scanContinuedUnicodes(next, collect);
  auto atom = (
    cnt == 1
    ? getCharAtom(chr)
    : sptrOf<TextAtom>(_latex.substr(_pos, n), _isMathMode)
  );
  _pos += n - 1;
  return atom;
}

```

src/Data/Convert/Texmacs/totm.cpp
```diff
@@ -144,6 +145,13 @@ tm_writer::write (string s, bool flag, bool encode_space) {
     for (i=0; i<n; i++) {
       char c= s[i];
       if ((c == ' ') && (!encode_space)) write_space ();
+      else if (get_preference ("tm format with utf8", "on") == "on" &&
```
darcy-shen, line 148: for GNU TeXmacs, we will switch it off by default

PR #314, opened 2022-09-15
[no ci] Development Guide on Windows/macOS/Linux

darcy-shen:
+ Unify the launcher script to `mogan.sh`
+ Developers' Guide ZH/EN
  + https://github.com/XmacsLabs/mogan/blob/da/000/docs/Develop_EN.md
  + https://github.com/XmacsLabs/mogan/blob/da/000/docs/Develop_ZH.md


darcy-shen:
I think I should rename `mogan.sh` back to `mogan`.

PR #324, opened 2022-09-27
Update cpp language parser-feature

luozhiya:
Add:
* nullptr
* nullptr_t
* byte
* char16_t
* char32_t
* char8_t
* int16_t
* int32_t
* int8_t
* uint16_t
* uint32_t
* uint8_t
* concept
* constexpr


darcy-shen:
Thanks for your contribution!

PR #341, opened 2022-10-09
option for TM format with UTF8 support (continued 2)

da-liii:


src/Data/String/converter.cpp
```diff
@@ -808,6 +808,16 @@ encode_as_utf8 (unsigned int code) {
   else return "";
 }
 
+string encode_iso_8859_as_utf8 (unsigned char c) {
```
Yufeng-shen, line 811: Any reference for this converter? Does it only work for ISO-8859-1? I'm asking because it seems that there are 16 different parts in the ISO-8859 family (https://en.wikipedia.org/wiki/ISO/IEC_8859)

darcy-shen: Let me learn more about ISO-8859 and then continue. It is now a stackoverflow oriented pull request.

darcy-shen:
Well, I found it too complicated for us to solve it in v1.1.1. Let me close it now.

darcy-shen:
With option `tm format with utf8` on, the tm format is actually a mix of UTF8 (only the CJK part) and ISO-8859.

PR #376, opened 2022-11-09
CMake: indent and format

darcy-shen:
+ Use 2 spaces for indenting
+ Put one space after `if`


jingkaimori:
mingw may fail to install occasionally, just rerun jobs to fix it

PR #375, opened 2022-11-07
Add shortcut, use utf-8 for file

ice1000:
Fix #317 I guess

PR #378, opened 2022-11-10
Add delete word shortcut

### LINKED ISSUE #317: Ctrl+backspace delete a word as in Emacs
Very useful!!

ice1000:
Fix #317 see also #375

### DIFF
--- a/TeXmacs/progs/generic/generic-edit.scm
+++ b/TeXmacs/progs/generic/generic-edit.scm
@@ -677,6 +677,14 @@
   (selection-set-end)
   (clipboard-cut "primary"))
 
+(tm-define (backward-kill-word)
+  (kbd-select traverse-left)
+  (kbd-delete))
+
+(tm-define (kill-word)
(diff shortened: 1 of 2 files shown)


darcy-shen:
Currently, CI only works for C++. You can use the prefix `[no ci]` in the commit message to disable CI.

darcy-shen (changes requested):
Here is the right place to maintain Emacs-style key-bindings.

And in emacs, the corresponding elisp function is `backward-kill-word` and `kill-word`.

https://github.com/XmacsLabs/mogan/blob/bc0051a366b58032cf8ce178fbcd367ab37f8993/TeXmacs/progs/generic/generic-kbd.scm#L266-L270

ice1000:
> And in emacs, the corresponding elisp function is backward-kill-word and kill-word.

I tried to grep these names but I didn't find them

darcy-shen (approved):
LGTM

PR #379, opened 2022-11-12
xmake: TEXMACS_VERSION should be 2.1.2 15m

darcy-shen:


darcy-shen:
I forgot to append the 15m to the commit message. And I just changed the merge settings, it will use pull request title and description as commit message later.

PR #383, opened 2022-11-16
xmake: dev guide on GNU/Linux 45m

darcy-shen:


jingkaimori:
package feature should be described.
```sh
xmake install mogan_install
./build/package/bin/mogan
```

darcy-shen:
This pull request only aim to describe about the dev guide. On GNU/Linux, packaging is different for different Linux distro.

darcy-shen:
On Debian, `xmake install mogan_install` will install to `/usr/local`.

darcy-shen:
Do you mean like windows:

```
xmake install -o build/package mogan_install
```
And then launch it via 
```
build/package/bin/mogan
```

PR #395, opened 2022-11-28
refactor(packages/macos): Cleaner Structure

pan93412:
## Description

Rewrite `packages/macos/package.sh` for better error handling, robust structure and self-explainable code.

## TODOs

- [x] Could we find a more elegant solution?

## Related

- CI: https://github.com/pan93412/mogan/actions/runs/3594111085/

### DIFF
--- a/.github/workflows/package-macos.yml
+++ b/.github/workflows/package-macos.yml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - 'v*'
+  workflow_dispatch:
 
 jobs:
   mac-package:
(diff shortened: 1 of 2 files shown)


darcy-shen:
Error handling should only be treated carefully on crucial path. We can use the same exit code with a error message.

pan93412:
> Error handling should only be treated carefully on crucial path. We can use the same exit code with a error message.

I think so. Let me refactor the script.

pan93412:
@darcy-shen Actually this script has some bugs (see CI), sorry for not marking it as “Draft” :(

I'm testing the patch, and I'll make another PR that fixes it.

PR #399, opened 2022-12-01
xmake: require sqlite3

darcy-shen:


jingkaimori (commented):
syntax ok, but why we must depend on sqlite3?

darcy-shen:
> syntax ok, but why we must depend on sqlite3?

The TeXmacs bib data tool depends on Sqlite3.

PR #413, opened 2022-12-06
Add support for macOS M1

pan93412:
- Add the Homebrew Qt path of macOS M1.
- Sign `Mogan.app` after `macdeployqt` as this is a mandatory requirement for M1.

CI: <https://github.com/pan93412/mogan/actions/runs/3626346588> (passed)

### DIFF
--- a/packages/macos/package.sh
+++ b/packages/macos/package.sh
@@ -44,6 +44,7 @@ copy_assets() (
 
 deploy_app() {
     macdeployqt Mogan.app -verbose=1 -dmg
+    codesign --force --sign - --deep Mogan.app
 }
 
 build_mogan_base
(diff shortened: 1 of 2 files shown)


packages/macos/package.sh
```diff
@@ -44,6 +44,7 @@ copy_assets() (
 
 deploy_app() {
     macdeployqt Mogan.app -verbose=1 -dmg
+    codesign --force --sign - --deep Mogan.app
```
darcy-shen, line 47: codesign should be applied before `macdeployqt`

pan93412: Oh, I'll fix it later.

darcy-shen: Well, it is a bit tricky. We need to invoke macdeployqt first but we should not create the dmg.

And then, we should apply codesign on the `macdeployqt`ed Mogan.app, and finally, generate the dmg.

darcy-shen: For Mogan v1.1.2, we will switch to xmake. xmake is much more user-friendly.

Here is the roadmap: http://forum.texmacs.cn/t/mogan-v1-1-2-roadmap/1174

pan93412:
I would switch to `Draft` and maybe we can check and fix it later.

darcy-shen:
`./packages/macos/package.sh` works fine on M1 (using xmake) now!

PR #464, opened 2022-12-31
fix ci artifacts

jingkaimori:


darcy-shen:
@jingkaimori I tried https://github.com/XmacsLabs/mogan/actions/runs/3812298741

This pull request did not fix the issue.

PR #471, opened 2022-12-31
Change version format on Windows

SpecterShell:
This PR changes the format of the version that the Windows installer writes to the uninstall registry key from `v1.1.2` to `1.1.2`, so it will align with the version string on Windows prior to 1.1.0 and other platforms (macOS, etc.). This will not affect current users.
image

By the way, the `VersionInfoTextVersion` property has been deprecated after Windows 98. It can be replaced by `VersionInfoVersion`.
https://jrsoftware.org/ishelp/index.php?topic=setup_versioninfotextversion

### DIFF
--- a/packages/windows/Xmacs.iss.in
+++ b/packages/windows/Xmacs.iss.in
@@ -9,7 +9,7 @@ DefaultGroupName=Mogan
 VersionInfoTextVersion=v@XMACS_VERSION@
 AppPublisher=XmacsLabs
 AppPublisherURL=http://github.com/XmacsLabs/mogan
-AppVersion=v@XMACS_VERSION@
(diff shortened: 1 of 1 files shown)


darcy-shen:
For AppVersion, here is the doc: https://jrsoftware.org/ishelp/topic_setup_appversion.htm

Thanks for this pull request!

darcy-shen (approved):

darcy-shen:
> By the way, the VersionInfoTextVersion property has been deprecated after Windows 98. It can be replaced by VersionInfoVersion.
> https://jrsoftware.org/ishelp/index.php?topic=setup_versioninfotextversion

Welcome to create another pull request to deprecate the VersionInfoTextVersion property.

PR #504, opened 2023-01-20
New icon for Mogan on macOS

mgubi:
I've added the new icon and also changed the icon in the documentation. For the moment I've only worked out it for MacOSX. This is the appeareance
new-mogan-128


darcy-shen:
Tested on macOS! It works, but the icon in the dash is still using the old icon. It is not a big problem. I will improve it when working on the icon change for Windows and Linux.

mgubi:
> Tested on macOS! It works, but the icon in the dash is still using the old icon. It is not a big problem. I will improve it when working on the icon change for Windows and Linux.

Yes, this happens to me too. I do not understand why. Possibly a problem with the cache of the Docker which is still retaining the old icon?

PR #510, opened 2023-01-22
[112] Chinese: Prohibition Rules for Line Start and Line End 4h30m

darcy-shen:
Implemented the basic rule of https://www.w3.org/International/clreq/#prohibition_rules_for_line_start_end
It may break other rules.

### DIFF
--- a/docs/devel/112.tm
+++ b/docs/devel/112.tm
@@ -1,11 +1,11 @@
 <TeXmacs|2.1.2>
 
-<style|tmdoc>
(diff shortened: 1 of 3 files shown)


YiqiXu:
The nice document also contains rules for English-Chinese bilingual mixed typesetting. Do you plan to implement those, too?

darcy-shen:
> Do you plan to implement those, too?

5 days left. Long term, the answer is yes. For v1.1.2, I do not know.

PR #530, opened 2023-02-19
fix build with c++20

PikachuHy:
add extra `string_view` to solve 
use of overloaded operator '==' is ambiguous (with operand types 'tree_label' and 'const char[]')

### DIFF
--- a/src/Kernel/Types/string.hpp
+++ b/src/Kernel/Types/string.hpp
@@ -38,6 +38,7 @@ class string {
   string (const char *s);
   string (const char *s, int n);
   inline char& operator [] (int i) { return rep->a[i]; }
+  inline char operator [] (int i) const { return rep->a[i]; }
   bool operator == (const char* s);
   bool operator != (const char* s);
   bool operator == (string s);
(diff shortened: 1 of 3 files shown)


src/Kernel/Types/string.hpp
```diff
@@ -116,4 +117,31 @@ class c_string {
...
+class string_view {
+public:
+  string_view(const char* data, int len): n(len), a(data) {}
+  inline char operator [] (int i) const { return a[i]; }
+private:
+  friend inline int N (string_view s);
+  int n;
+  const char* a;
+};
+inline string_view operator""_sv(const char* data, std::size_t len) {
```
darcy-shen, line 129: suffixing a string literal with `_sv` looks good.

Is it a good idea to suffixing a string literal with `_u` to indicate it is a unicode string?

src/Plugins/Qt/qt_chooser_widget.cpp
```diff
@@ -175,7 +175,7 @@ qt_chooser_widget_rep::set_type (const string& _type)
     nameFilter = to_qstring (translate
                              (as_string (call ("format-get-name", _type))
                               * " file"));
-  } else if (_type == "image") {
+  } else if (_type == "image"_sv) {
```
darcy-shen, line 178: is string_view a solution for equality between `const string&` and `string` in C++ 20?

Is there a better solution without needing to append the extra `_sv` suffix?

darcy-shen: will `_type == string("image")` solve the problem?

PikachuHy: yes

PikachuHy: > is string_view a solution for equality between const string& and string in C++ 20?

no, I just want to build success with c++20

the `_type` is string, and the `"image"` is `char[6]`.
however, the operator `==` can not handle the case, weird

PikachuHy:
the windowsbuild ci failed ...

src/Kernel/Types/string.hpp
PikachuHy: the operator `_sv` is just a grammar sugar, maybe

PikachuHy:
https://github.com/XmacsLabs/mogan/pull/533

PikachuHy:
https://github.com/XmacsLabs/mogan/pull/532

PR #536, opened 2023-03-16
Eliminate warnings caused by action of mingw setup.

jingkaimori:
Main branch of action to set up mingw is egor-tensin/setup-mingw


darcy-shen:
No doc on before and after, we'd better add a doc to describe about it

PR #541, opened 2023-03-24
split ci workflows

jingkaimori:


darcy-shen:
https://github.com/XmacsLabs/mogan/actions

The name `Build via xmake` should be changed into:
+ Build via xmake on macOS
+ Build via xmake on Linux
+ Build via xmake on Windows

PR #544, opened 2023-03-25
upgrade pdfhummus in xmake config to avoid export failure

jingkaimori:


docs/devel/127.tm
```diff
@@ -0,0 +1,41 @@
...
+  <\itemize>
+    <item>Reporter: <hlink|斩夜神|https://gitee.com/dhuzf> on
+    <hlink|Gitee|https://gitee.com/XmacsLabs/mogan/issues/I6B6AH>
+
+    <item>Owner: jingkaimori
+  </itemize>
+
+  <section|Description>
+
+  本次提交更改了xmake.lua文件，使用xmake构建时，将安装版本较新的PDFHummus，代替内嵌的版本。本次提交不会更改cmake的行为，使用cmake构建的版本仍然存在bug。
```
darcy-shen, line 19: AI translation might be good enough in the near future. I think it is ok to use Chinese here.

jingkaimori: maybe a better method is develop a macro to display multilanguage text, which is implemented by mediawiki.

darcy-shen:
After we complete the wasm build in xmake, we should deprecate the cmake support.

PR #580, opened 2023-04-23
[170] Use XDG_CACHE_HOME on GNU/Linux 4h

darcy-shen:


TeXmacs/progs/texmacs/texmacs/tm-tools.scm
```diff
@@ -70,7 +70,7 @@
   (:synopsis "Clear font cache under TEXMACS_HOME_PATH.")
   (map system-remove
     (list
-      "$TEXMACS_HOME_PATH/system/cache/font_cache.scm"
+      "$XDG_CACHE_HOME/app.mogan.editor/font_cache.scm"
```
jingkaimori, line 73:
```suggestion
      (string-append (url->system (get-texmacs-cache-path)) "font_cache.scm")
```
why use environment variable explicitly?

TeXmacs/progs/kernel/texmacs/tm-plugins.scm
```diff
@@ -333,7 +333,7 @@
 
 (define-public reconfigure-flag? #t)
 (define plugin-loaded-setup? #f)
-(define plugin-cache "$TEXMACS_HOME_PATH/system/cache/plugin_cache.scm")
+(define plugin-cache (string-append (url->system (get-texmacs-cache-path)) "/plugin_cache.scm"))
```
jingkaimori, line 336: maybe url-append is better

PR #659, opened 2023-05-03
[3_23] Removing bundled nowide

jingkaimori:
This pr should be reviewed, because file.cpp changes, thus I use procedures in dirent.h directly.


darcy-shen:
> This pr should be reviewed, because file.cpp changes, thus I use procedures in dirent.h directly.

The `file.cpp` will be re-implemented later. I planned to use tbox to re-impl it.

PR #688, opened 2023-05-26
[14_1] add unit tests for tree

pokyux:


tests/Kernel/Types/tree_test.cpp
```diff
@@ -19,6 +19,9 @@ private slots:
   void test_is_atomic ();
   void test_is_tuple ();
   void test_is_concat ();
+  void test_is_compound();
```
darcy-shen, line 22: please run `bin/format`, before run `bin/format`, please install clang-format via pacman in msys2.

pokyux: Done.

PR #693, opened 2023-06-01
[14_2] add unit tests for rectangle

Oyyko:
add unit tests for rectangle


tests/Graphics/Types/rectangle_test.cpp
```diff
@@ -0,0 +1,58 @@
+
+/******************************************************************************
+ * MODULE     : tree_test.cpp
```
darcy-shen, line 3: wrong filename

PR #706, opened 2023-06-10
[15_3] Separate gen_l1 from glue_basic

darcy-shen:


src/Scheme/Scheme/glue.cpp
```diff
@@ -589,6 +589,7 @@ string recognize_glyph (array_array_array_double gl);
 #include "new_style.hpp"
 #include "persistent.hpp"
 
+#include "gen_l1.cpp"
```
jingkaimori, line 592:
```suggestion
#include "gen_l1.cpp"
#include "glue_tree.cpp"
```

src/Scheme/Scheme/glue.cpp
```diff
@@ -599,6 +600,7 @@ initialize_glue () {
   
   initialize_glue_l1 ();
   initialize_glue_l2 ();
+  initialize_gen_l1 ();
```
jingkaimori, line 603:
```suggestion
  initialize_gen_tree ();
  initialize_gen_l1 ();
```

jingkaimori (changes requested):

src/Scheme/Scheme/glue.cpp
darcy-shen: I'm wondering how it pass the ci

jingkaimori: because ci does not test whether these function is registered successfully

PR #768, opened 2023-06-29
[23_6] Task: Feature, Add more arrow types

Oyyko:


devel/23_6.tm
```diff
@@ -0,0 +1,45 @@
...
+    <item*|Github issue>NA
+  </description>
+
+  <section|Description>
+
+  <subsection|Feature Description>
+
+  Add more arrows. For example:
+
+  <image|figures/23_6_1.png|269pt|111pt||>
```
darcy-shen, line 26: just put the image inside the tm doc

Oyyko: OK.

Oyyko: Fixed.

PR #773, opened 2023-07-03
[8_1] xmake: merge mogan_install into mogan

darcy-shen:


darcy-shen:
Documentation will be updated in a separated pull request.

PR #775, opened 2023-07-04
[23_11] When pasting an object, it is necessary to keep a certain distance from the original object.

Oyyko:


Oyyko:
Still a draft.
Current progress: The pasted graphic will be placed in a new position according to the position of the graphic on the page. If the graphic is at the upper left corner, the pasted graphic will shift to the lower right. and so on.
I want to ask some questions.
1. Where should I put helper functions? Should they all be written into one function definition? Or place it in a new file.
2. I did not find some functions. For example to get the x-coordinate of a point. The version I implemented myself is as follows:
```scheme
(define (get-x pt)
    (s2f (cadr pt)))
```
I'm wondering if I can refactor some code to use this new function. For example in the original code:
```scheme
(define (translate-point x y)
  (lambda (o)
    (if (match? o '(point :%2))
	`(point ,(f2s (+ x (s2f (cadr o)))) ,(f2s (+ y (s2f (caddr o)))))
        o)))
```
The original author did not define these helper functions, but wrote them directly. This can get confusing.
3. Please advise on my coding style.

PR #780, opened 2023-07-08
[8_1] add icon for mogan under windows

jingkaimori:
rc was introduced from separate target, and it will be enabled only under windows environment.

### DIFF
--- a/xmake.lua
+++ b/xmake.lua
@@ -656,6 +656,19 @@ option("libdl") do
     add_links("dl")
 end
 
+target("windows_icon") do
+    set_enabled(is_plat("mingw"))
+    set_version(XMACS_VERSION)
+    set_kind("object")
+    add_configfiles("packages/windows/resource.rc.in", {
+        filename = "resource.rc"
+    })
+    add_configfiles("packages/windows/TeXmacs.ico", {
+        onlycopy = true
+    })
+    add_files("$(buildir)/resource.rc")
+end
+
 target("mogan") do 
     set_version(XMACS_VERSION)
     if is_plat("macosx") then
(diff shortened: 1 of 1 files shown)


Yufeng-shen (approved):
looks good to me

xmake.lua
```diff
@@ -656,6 +656,20 @@ option("libdl") do
...
+    target("windows_icon") do
+        set_version(XMACS_VERSION)
+        set_kind("object")
+        add_configfiles("packages/windows/resource.rc.in", {
+            filename = "resource.rc"
+        })
+        add_configfiles("packages/windows/TeXmacs.ico", {
+            onlycopy = true
+        })
+        add_files("$(buildir)/resource.rc")
```
darcy-shen, line 669:
```
warning: .\xmake.lua:669: cannot match add_files("build/resource.rc") in target(windows_icon)
```

darcy-shen: https://github.com/xmake-io/xmake/issues/3229
another problem is: `error: ...mdir\core\sandbox\modules\import\core\base\scheduler.lua:56: cannot get program for mrc`

darcy-shen: `rc.exe`是需要安装Windows SDK，这对于开发者来说是不需要的。这块应该放在打包的阶段

darcy-shen (changes requested):
windows_icon should be build when packaging

xmake.lua
darcy-shen: https://github.com/xmake-io/xmake/issues/3942

不需要安装 Windows SDK，只需要把mingw加到path里面就可以了

darcy-shen (approved):
LGTM now

PR #784, opened 2023-07-10
[8_1] deploy cxx runtime from qt

jingkaimori:
since version of mingw used to build mogan may differs from version of Qt, tell windeployqt to use lib from mingw may break ABI compability.


darcy-shen (approved):
awesome

PR #810, opened 2023-07-20
[5_12] Enable search for the website

xiaoliyuhh:
Added search function


darcy-shen:
You can learn about topic branch: https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows

PR #880, opened 2023-08-03
[46_1] change default value of par-no-first

jingkaimori:


jingkaimori:
If `par-no-first` should not be toggled by user, it is not necessary to read its initial value, thus changing default value is meaningless.

PR #889, opened 2023-08-04
[23_21] Ban jump in text/math at of graphics area

Oyyko:


darcy-shen:
Tested, and it works.

Please write a detailed doc in 23_21.tm, if you think writing in English is a bit hard, feel free to write the doc in Chinese.

PR #892, opened 2023-08-05
[29_1] add Export pdf with tm option as experimental feature

tangdouer1005:


TeXmacs/progs/texmacs/menus/preferences-widgets.scm
```diff
@@ -800,6 +800,10 @@
           (toggle
            (set-boolean-preference "gui:new bibliography dialogue" answer)
            (get-boolean-preference "gui:new bibliography dialogue")))
+        (meti (hlist // (text "Export PDF tm attachment"))
```
darcy-shen, line 803: Export PDF with tm attachment

tangdouer1005: I fixed this syntax error

PR #899, opened 2023-08-06
[23_8] Auto Unselect everything when entering the move mode in the graphics mode.

Oyyko:


darcy-shen (approved):
Tested, it works fine.

PR #911, opened 2023-08-08
[23_20] Fix a bug that would cause an infinite loop

Oyyko:
In addition, there are still some issues with the functionality of 23_20. I have browsed the code repository, but I couldn't figure out why the menu button sometimes doesn't change to a dark color. Given the work schedule of OSPP, I will temporarily suspend this task.

Currently, as long as users follow the normal procedure, there won't be any inconsistencies. However, if a user is in the circle drawing mode and clicks the circle button again, it will enter the mode for setting properties. At this point, if the user clicks the circle button again, it will return to the circle drawing mode. The mode switching works fine. However, the circle button does not change to a dark color.

### DIFF
--- a/TeXmacs/progs/graphics/graphics-main.scm
+++ b/TeXmacs/progs/graphics/graphics-main.scm
@@ -856,9 +856,8 @@
 
 (tm-define (graphics-set-mode val)
   (:check-mark "v" graphics-mode-has-value?)
-  (if (and
-        (equal? (graphics-mode) val)
-        (not (equal? (graphics-mode) '(group-edit edit-props))))
+  (if (and (equal? (graphics-mode) val)
+           (not (equal? (graphics-mode) '(group-edit edit-props))))
     (graphics-set-mode '(group-edit edit-props))
     (begin
       (if (== val '(group-edit move))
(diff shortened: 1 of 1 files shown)


TeXmacs/progs/graphics/graphics-main.scm
```diff
@@ -856,7 +856,9 @@
 
 (tm-define (graphics-set-mode val)
   (:check-mark "v" graphics-mode-has-value?)
-  (if (equal? (graphics-mode) val)
+  (if (and
```
darcy-shen, line 859: code formatting of `and` should following the code formatting below

TeXmacs/progs/graphics/graphics-main.scm
```diff
@@ -856,7 +856,9 @@
...
+        (equal? (graphics-mode) val)
+        (not (equal? (graphics-mode) '(group-edit edit-props))))
     (graphics-set-mode '(group-edit edit-props))
```
darcy-shen, line 862: any good way to use unit test to make sure that there is no infinite loop?

Oyyko: According to the Halting Problem, it is not possible to use unit testing to guarantee the absence of infinite loops.

TeXmacs/progs/graphics/graphics-main.scm
Oyyko: Do you think it is feasible to fix like this?

TeXmacs/progs/graphics/graphics-main.scm
darcy-shen: In practice, infinite loops will lead to timeout of CI, and in this way, we will find the error.

PR #914, opened 2023-08-08
[50_1] refactor init_app

Oyyko:


src/Texmacs/Server/tm_server.cpp
```diff
@@ -98,12 +97,10 @@ tm_server_rep::tm_server_rep (): def_zoomf (1.0) {
...
   gui_interpose (texmacs_interpose_handler);
   set_wait_handler (texmacs_wait_handler);
-  if (is_none (tm_init_file))
-    tm_init_file= "$TEXMACS_PATH/progs/init-texmacs-s7.scm";
+  
   if (is_none (my_init_file))
     my_init_file= "$TEXMACS_HOME_PATH/progs/my-init-texmacs.scm";
   bench_start ("initialize scheme");
-  if (exists (tm_init_file)) exec_file (tm_init_file);
   if (exists (my_init_file)) exec_file (my_init_file);
```
darcy-shen, line 104: This is the init file provided by the user, it should also be put in init_app

src/Mogan/Research/research.cpp
```diff
@@ -472,7 +493,7 @@ TeXmacs_main (int argc, char** argv) {
   if (DEBUG_STD) debug_boot << "Starting server...\n";
   { // opening scope for server sv
     server sv;
-
+    init_app (now_app);
```
darcy-shen, line 496: init_app should be placed in the constructor of class server

and we can initialize in that way:
```
server sv(App::Draw)
```

src/Mogan/Research/research.cpp
```diff
@@ -75,6 +76,26 @@ static QTMApplication*     qtmapp    = NULL;
 static QTMCoreApplication* qtmcoreapp= NULL;
 #endif
 
+/******************************************************************************
+ * Init applications
+ ******************************************************************************/
+ 
+enum class App { RESEARCH, DRAW };
```
darcy-shen, line 83: `git grep enum`, you will find that the codestyle of enum in texmacs, in texmacs, we should use `enum class app`

src/Mogan/Research/research.cpp
```diff
@@ -75,6 +76,26 @@ static QTMApplication*     qtmapp    = NULL;
...
+constexpr App now_app= App::RESEARCH;
+
+void
+init_app (App app) {
+  if (is_none (tm_init_file)) {
+    if (app == App::DRAW) {
+      tm_init_file= "$TEXMACS_PATH/progs/init-draw.scm";
+    }
+    else if (app == App::RESEARCH) {
+      tm_init_file= "$TEXMACS_PATH/progs/init-texmacs-s7.scm";
```
darcy-shen, line 93: how about rename it to `init-research.scm`

And the string `research`, `draw` could be generated via C++ enum.

See https://en.cppreference.com/w/cpp/language/enum

src/Mogan/Research/research.cpp
```diff
@@ -52,7 +53,7 @@ void mac_fix_paths ();
 
 extern bool char_clip;
 
-extern url    tm_init_file;
```
darcy-shen, line 55: Note that there is a TeXmacs server mode. I guess that's why it is extern here.

darcy-shen (changes requested):
Please make minimum changes to the codebase to support app type when initializing.

src/Mogan/Research/research.cpp
Oyyko: I think that the reflection capabilities of C++ are currently insufficient to support the conversion of enum to texmacs string or C-style string. Additionally, if we were to use the texmacs string type here, we would be unable to achieve compile-time computing. But with C-style string, they cannot be added together in compile-time.

Oyyko:
Refactored. Please check if it is elegant.

darcy-shen (approved):

PR #915, opened 2023-08-08
[50_1] New Entry for Mogan Draw

Oyyko:


xmake.lua
```diff
@@ -485,6 +485,228 @@ if is_plat("mingw") then
     end
 end
 
+target("draw") do
```
darcy-shen, line 488: just ignore the code for mingw and macos

it should only be enabled on Linux and wasm

darcy-shen: Focusing on wasm and linux, will make the build definition much more simpler.

darcy-shen:
It depends on #914, in general, it is ok. To make it work on WASM, we just need to focus on Linux and wasm, based on the restricted platform fact, we can just remove the mingw and macOS related dead code.

xmake.lua
darcy-shen: We can use set_enabled here: https://xmake.io/#/manual/project_target?id=targetset_enabled

eg:
```
set_enabled(is_plat("linux" or is_plat("wasm"))
```

Oyyko:
This branch is a little mess... I will close this PR and create a new branch for the new entry of mogan draw.

PR #935, opened 2023-08-14
[29_2] load tm attachement in pdf

tangdouer1005:


Oyyko (commented):
This file maybe need a copyright header. Besides, usually there is a blank line at the end of a file.

src/Plugins/Pdf/pdf_hummus_get_attachment.cpp
Oyyko, line 1: This file maybe need a copyright header. Besides, usually there is a blank line at the end of a file.

tangdouer1005: you are right

tangdouer1005: I didn't notice this before, why is there a blank line at the end of each file？

Oyyko: https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

Oyyko: There is also the point of keeping diff history. If a file ends without a newline character, then adding anything to the end of the file will be viewed by diff utilities as changing that last line (because \n is being added to it).

This could cause unwanted results with commands such as git blame.

tangdouer1005: I got, thank you

PR #943, opened 2023-08-18
[18_5] Improve document of tm-define

jingkaimori:
http://forum.texmacs.cn/t/how-to-use-options-in-tm-define/1527


jingkaimori:
this document will not be contributed to mogan, submitting pr to upstream

darcy-shen:
You can request SVN permission from Joris. I think you are already a GNU TeXmacs developer.

PR #1006, opened 2023-09-09
[32_8] set minimal xmake version to 2.8.2

darcy-shen:
Because we are using `add_forceincludes`.
https://xmake.io/#/manual/project_target?id=targetadd_forceincludes


Charonxin (approved):
LGTM

PR #1007, opened 2023-09-09
[53_7] Replace (url-test? u "c") with url-exists?

### LINKED ISSUE #1005: Failed to Save as a tm file
Here is the error message related code:
```
(define (load-buffer-check-permissions name opts)
  ;;(display* "load-buffer-check-permissions " name ", " opts "\n")
  (with vname `(verbatim ,(url->system name))
    (cond ((and (not (url-test? name "f")) (not (url-test? name "c")))
           (with msg `(concat "The file " ,vname
                              " cannot be loaded or created")
```

darcy-shen:
`url-test? u "c"` returns `true` when u does not exist before, we change the semantics recently.

close #1005


Charonxin (approved):
Cool!

PR #1008, opened 2023-09-09
[53_7] Use lolly 1.2.4

paradisuman:


Charonxin (changes requested):
tidy :

- https://github.com/XmacsLabs/mogan/blob/branch-1.2/src/System/config_l3.h.xmake#L14-L15
- https://github.com/XmacsLabs/mogan/blob/branch-1.2/src/System/config_l3.h.xmake#L14-L15

src/Scheme/S7/s7_tm.cpp
```diff
@@ -16,16 +16,14 @@
 #include "object_l1.hpp"
 #include "object_l2.hpp"
 #include "tm_debug.hpp"
+#include "tbox/tbox.h"
```
darcy-shen, line 19: mogan里面不能直接用 tbox ，可以用lolly

darcy-shen:
Sorry, I have to take it over to publish lolly 1.2.4 now for wasm version of Mogan Research.

PR #1015, opened 2023-09-10
[53_9] Use lolly 1.2.4

darcy-shen:


Charonxin:
cc: 

- [ ] https://github.com/XmacsLabs/mogan/blob/branch-1.2/src/System/config_l3.h.xmake#L13-L15
- [ ] https://github.com/XmacsLabs/mogan/blob/branch-1.2/src/System/config.h.xmake#L16-L22

PR #877, opened 2023-08-02
Revert "[23_12] Move Exit Graphics Button to right"

YiqiXu:
Reverts XmacsLabs/mogan#873

https://github.com/XmacsLabs/mogan/pull/864#issuecomment-1660788096 The issues are not yet resolved.

PR #1036, opened 2023-09-16
[23_12] Revert the location of Exit Graphics Button

Oyyko:

Revert #873 . See #877 also.

- [x] Revert the location of Exit Graphics Button from Right to Left.


darcy-shen:
I do not think we need to revert this UI change. I will not approve this pull request.

PR #1071, opened 2023-09-22
[48_11] wasm: remove toolbar button with popup menu, solve the dialog crash

da-liii:
Port #451 to branch-1.2


darcy-shen:
Tried, it does not work on branch-1.2, and in branch-1.1, it is not a complete fix.

darcy-shen:
Should be closed by #1143

PR #1086, opened 2023-09-24
[53_12] use lolly 1.2.7

paradisuman:

use lolly 1.1.7

- change lolly version to 1.2.7

### DIFF
--- a/misc/xmake/packages.lua
+++ b/misc/xmake/packages.lua
@@ -35,7 +35,7 @@ package("lolly")
         add_deps("libcurl")
     end
 
-    add_versions("v" .. LOLLY_VERSION, "11633c9c4929d2f7d3e36bb75deab2b3a7aa3c24")
+    add_versions("main", "ddbea7e3ca89644578af735b913d34c239241350")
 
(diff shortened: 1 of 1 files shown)


jingkaimori:
maybe tag 1.2.7 of lolly should be update

misc/xmake/packages.lua
```diff
@@ -35,7 +35,7 @@ package("lolly")
         add_deps("libcurl")
     end
 
-    add_versions("v" .. LOLLY_VERSION, "11633c9c4929d2f7d3e36bb75deab2b3a7aa3c24")
+    add_versions("main", "abe2055a1a1778d2949dc70f697efa016487b290")
```
Charonxin, line 38: change `main` to release tag please

darcy-shen: it is ok, we have to test on main first and then bump the version

darcy-shen: And it should be marked as draft, this is the problem.

darcy-shen (approved):
LGTM

PR #1095, opened 2023-09-26
[29_6] Fix pdf plugin on windows

tangdouer1005:
in scheme func texmacs->pdf temporary url fail to generate on windows, 

Specifically, the character ';' will appear in the string converted into the new URL generated instead of the expected ':', causing the file to fail to open.

replace `(tem-pdf (string->url (string-append (url->string tem-dir) "/tem.pdf")))` with `(tem-pdf (url-append tem-dir "/tem.pdf"))`


Charonxin (approved):
Why are you so productive?

tangdouer1005:
> Why are you so productive?

because ddl is coming！:(

PR #1096, opened 2023-09-26
[29_6] add error handling for pdf->texmacs

tangdouer1005:
Incorrectly loading a pdf without attachments and detaching the attachment returns false. In this case, no error handling is performed and the entire program will crash. Add error handling to prevent crashes.

### DIFF
--- a/plugins/pdf/progs/init-pdf.scm
+++ b/plugins/pdf/progs/init-pdf.scm
@@ -14,11 +14,11 @@
 
 (define (texmacs->pdf x . opts)
   (let* ((tem-dir (url-temp-dir))
-          (tem-pdf (url-append tem-dir "/tem.pdf"))
-          (tem-tm (url-append tem-dir "/tem.tm"))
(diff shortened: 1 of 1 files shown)


plugins/pdf/progs/init-pdf.scm
```diff
@@ -29,26 +29,24 @@
...
     (switch-to-buffer cur)
     (buffer-close buf))
-  ;;attachment-list need be generated
   (if (pdf-make-attachments tem-pdf url-list tem-pdf)
-    (display* "pdf-make-attachments return true\n")
-    (display* "pdf-make-attachments return false\n"))
-  (string-load tem-pdf)))
+    (string-load tem-pdf)
+    (begin
+      (notify-now "Can not extract attachments from PDF")
```
darcy-shen, line 35: wrong error message

darcy-shen (approved):
LGTM

darcy-shen:
Verified on Windows, tried to open a PDF without tm attachment, it works fine.

PR #1110, opened 2023-09-28
[49_5] use qt.console instead of -mconsole

jingkaimori:

Simplify xmake.lua and avoid setting linker flag manually.

see also https://github.com/xmake-io/xmake/discussions/3910

- [x] Remove link option `-mconsole`
- [x] Simplify checking for console display, now checking on windows also.

### DIFF
--- a/xmake.lua
+++ b/xmake.lua
@@ -804,10 +804,10 @@ function add_target_research_on_others()
         set_optimize("smallest");
     end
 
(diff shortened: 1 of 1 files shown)


darcy-shen:
> now checking on windows also.

It confused me for a while, thanks! Now, I can switch to msvc on Windows for daily development!

PR #1133, opened 2023-10-02
[65_2] Use Qt 6.5.1 on windows

darcy-shen:


jingkaimori:
https://github.com/XmacsLabs/mogan/pull/480/files#diff-f05b44659458c38457572d597b36fcef2e1a7245e0c7323fe99f979930bef01dL510-R514
`QRegExp` should be replaced by `QRegularExpression` in this case.

darcy-shen:
Let me disable mingw ci temporarily.

PR #1143, opened 2023-10-03
[48_11] Fix drop down menu on toolbars

darcy-shen:


darcy-shen:
Here is the reason of failed CI on wasm:
The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.

PR #1147, opened 2023-10-03
[57_2] Remove ispell from libmogan

Charonxin:

Just delete Ispell plugin or move to another dir?

- [x] comment ispell related code


xmake.lua
```diff
@@ -500,7 +500,7 @@ target("libmogan") do
     add_files(plugin_freetype_srcs)
     add_files(plugin_database_srcs)
     add_files(plugin_ghostscript_srcs)
-    add_files(plugin_ispell_srcs)
+    -- add_files(plugin_ispell_srcs)
```
jingkaimori, line 503: better method is setting `plugin_ispell_srcs` to an empty list

jingkaimori (changes requested):

xmake.lua
darcy-shen: > better method is setting plugin_ispell_srcs to an empty list

No. mutable variable is not a good practice

jingkaimori: > mutable variable is not a good practice

My suggestion is to change definition of `plugin_ispell_srcs`:
```lua
    local plugin_ispell_srcs = {}
    -- .....
    add_files(plugin_ispell_srcs)
```

darcy-shen: 这样也不合理，明明是 plugin_ispell_srcs，但是里面没有源代码文件

PR #1149, opened 2023-10-04
[57_5] Remove updater plugin from libmogan

darcy-shen:


jingkaimori:
this pr should be close because nsis installer needs sparkle to update automatically

PR #1150, opened 2023-10-04
[48_13] xmake: separate mogan code from xmake.lua

darcy-shen:


xmake/code.lua
```diff
@@ -0,0 +1,95 @@
...
+-- DESCRIPTION : Xmake config file for Mogan Code
+-- COPYRIGHT   : (C) 2022-2023  Darcy Shen
+--
+-- This software falls under the GNU general public license version 3 or later.
+-- It comes WITHOUT ANY WARRANTY WHATSOEVER. For details, see the file LICENSE
+-- in the root directory or <http://www.gnu.org/licenses/gpl-3.0.html>.
+
+function add_target_code()
+    set_languages("c++17")
+    set_version(XMACS_VERSION, {build = "%Y-%m-%d"})
```
jingkaimori, line 13: Why use global variable here? Use function parameter instead

xmake.lua
```diff
@@ -274,10 +274,10 @@ else
...
 local RUN_ENVS = {TEXMACS_PATH=path.join(os.projectdir(), "TeXmacs")}
 
-local DEVEL_VERSION = TEXMACS_VERSION
-local DEVEL_RELEASE = 1
-local STABLE_VERSION = TEXMACS_VERSION
-local STABLE_RELEASE = 1
+DEVEL_VERSION = TEXMACS_VERSION
+DEVEL_RELEASE = 1
+STABLE_VERSION = TEXMACS_VERSION
+STABLE_RELEASE = 1
```
jingkaimori, line 280: same as below

jingkaimori (changes requested):

xmake.lua
jingkaimori: can this part be reverted?

xmake.lua
```diff
@@ -258,7 +274,7 @@ target("libkernel_l3") do
 end
 
 
-local XMACS_VERSION="1.2.0"
+XMACS_VERSION="1.2.0"
```
jingkaimori, line 277: here

xmake.lua
```diff
@@ -157,7 +173,7 @@ end
 -- Library: L3 Kernel
 --
 set_configvar("QTTEXMACS", 1)
-local CONFIG_USER = "XmacsLabs"
+CONFIG_USER = "XmacsLabs"
```
jingkaimori, line 176: and here

jingkaimori (changes requested):

xmake.lua
darcy-shen: cannot

darcy-shen: fail to config if it is local var

jingkaimori: so these vars should be moved to above `add_tm_configure`, or passed into this function as parameter

darcy-shen: let's keep it in this pr, this pr is for separate mogan code from xmake.lua, improvements on tm_configure could be done in other pr

darcy-shen: I agree that your opinion is a good idea, but let us move on rapidly if the changes is not harmful

jingkaimori (approved):
LGTM, but please fix definition of several variable after merge

PR #1154, opened 2023-10-04
[64_1] fix for de-embedding of images with unicode filename

jingkaimori:

fix cork encoding of filename in "save image as..." dialog

- [x] fix


TeXmacs/progs/generic/embedded-edit.scm
```diff
@@ -37,13 +37,13 @@
 
 (tm-define (embedded-suffix t)
```
darcy-shen, line 38: please add unit tests for these two tm-defined interface

jingkaimori: The second interface relies on `(current-directory)`, how to test it stably?

darcy-shen: should use integration test but not unit test

jingkaimori: unit test added @darcy-shen

darcy-shen (approved):
Verified locally via `xmake run 64_1`

PR #1170, opened 2023-10-05
[29_7] remove dead code

tangdouer1005:
In project 29, it is necessary to add conversion methods for pdf format and tm format, but the conversion method from tm to pdf is actually a pdf export method, so the method implemented here is useless

### DIFF
--- a/devel/29.tm
+++ b/devel/29.tm
@@ -57,6 +57,14 @@
       Tangdouer
     </cell>|<\cell>
       Add conversion mechanism between pdf and tm
+    </cell>>|<row|<\cell>
+      Chore
+    </cell>|<\cell>
+      29_7
+    </cell>|<\cell>
+      Tangdouer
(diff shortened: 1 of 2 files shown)


devel/29.tm
```diff
@@ -57,6 +57,14 @@
       Tangdouer
     </cell>|<\cell>
       Add conversion mechanism between pdf and tm
+    </cell>>|<row|<\cell>
```
jingkaimori, line 60: update of project manage document should be submit in separate pr, thus changes to source code can be extracted and ported to upstream texmacs project conveniently.

jingkaimori (changes requested):

devel/29.tm
tangdouer1005: ok, i had submitted a doc pr before this. i will modify it

PR #1253, opened 2023-10-15
Add cd for draw wasm

Oyyko:

Add CD for mogan draw wasm.

- [x] change xmake.lua and draw.lua for draw
- [x] Add CD

### DIFF
--- a/.github/workflows/cd-gh-pages-for-draw.yml
+++ b/.github/workflows/cd-gh-pages-for-draw.yml
@@ -1,0 +1,72 @@
+name: Github Pages for Mogan Draw
+
+on:
+  push:
+    branches: [ branch-1.2 ]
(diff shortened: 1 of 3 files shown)


darcy-shen:
Let me create a new repository  called draw.mogan.app, because it is a trivial task and requires some persimission, I will help you complete it.

Oyyko:
OK. In this PR, I suppose mogan draw runs on https://xmacslabs.github.io/mogan/draw/draw.html. It would be better if we can run it on draw.mogan.app.

darcy-shen:
There are two github action operating on the same branch `gh-pages`, it is a problem. And we should not keep gh-pages in this repo. It will increase the network payload when we git fetch from github.

PR #1254, opened 2023-10-16
[27_1] Draw Sector

Oyyko:
Add Draw Sector Function.
- [x] Add Draw Sector in menu
- [x] Implement Draw Sector (with fill color correctly)
- [ ] get a sector icon for menu
- [ ] follow #1255


Oyyko:
Still have problems when drawing "fat" sectors.

Oyyko:
It's hard to fix because the draw method is not designed for the OOP(in scheme, we got no Object concepts). Global variables can get entangled with each other. Try changing the sector function to return
```scheme
(if (> (points-cross-product-k vec-c-p vec-c-q) 0)
        `(superpose
          (std-arc ,c ,p ,m)
          (std-arc ,c ,m ,x)
          (with "line-width" "0ln"
              "color" ,(graphics-get-property "gr-fill-color")
            (line ,p ,c ,x ,m ,p)))
        `(superpose    
        (line ,p ,c ,x)
        (arc ,p ,m ,x)))))
```
But `,(graphics-get-property "gr-fill-color")` is strange here. It should be computed and never change. However, when I change the "gr-fill-color" in menu. The sectors which are already drawed also changed their color.

Oyyko:
I think it is unreasonable to implement the fan shape as a new shape like the ellipse. Because mogan uses parametric equations to describe graphics. The fan shape is composed of two line segments and an arc. It would be very strange to write a parametric equation to draw the fan shape.

mgubi:
> It looks like mogan will recalculate `,(graphics-get-property "gr-fill-color")` repeatedly, but I dont know why. As far as I know, it should not be recalculated. For example:

because the scheme procedure is called every time the texmacs macro is evaluated, which happens every time the document is typeset. you cannot change the color in the way you do. I think you need to get the value of that variable from the current TeXmacs environment (not scheme environment), instead that from the UI. Maybe with "(value  "gr-fill-color")" or something like this.

Oyyko:
> > It looks like mogan will recalculate `,(graphics-get-property "gr-fill-color")` repeatedly, but I dont know why. As far as I know, it should not be recalculated. For example:
> 
> because the scheme procedure is called every time the texmacs macro is evaluated, which happens every time the document is typeset. you cannot change the color in the way you do. I think you need to get the value of that variable from the current TeXmacs environment (not scheme environment), instead that from the UI. Maybe with "(value "gr-fill-color")" or something like this.

Thanks for your reply! So now my question is how to create an object(line, arc or something similar) in graphics mode with a color defined by programmers in code (not by users in texmacs).

Oyyko:
Maybe I can finish following https://github.com/XmacsLabs/mogan/pull/1255 at a new PR.

Oyyko:
Now problems:

- there is a very thin line in the sector, if you use a 4k-monitor you can see it. A little bit affecting the appearance
- no sector icons

darcy-shen:
> there is a very thin line in the sector, if you use a 4k-monitor you can see it.

Work fines to me.

darcy-shen:
For icon, we need to draw by our own. The icons in geogebra is nice, but it is non-free.

image

We can make our own icon in GNU TeXmacs style.

darcy-shen:
three points arc is not a good name, we may follow the name pattern adopted in GeoGebra.

PR #1425, opened 2023-12-09
[74] add document of underline extending bug

jingkaimori:

add document of underline extending bug, and related svn commits


darcy-shen:
In the undocumented new project management rule, all devel docs should start with `[num]` but not `[num_num]`.

PR #1432, opened 2023-12-09
[74_1] add integrated test for 74_1

jingkaimori:

This test is hard to be execute automatically, thus provide a test document

- [x] add test document


darcy-shen (approved):
LGTM

PR #1438, opened 2023-12-11
[65_8] Refactor: use locase and is_upper

darcy-shen:


jingkaimori:
Can this patch works on non-english keyboard?

darcy-shen:
> Can this patch works on non-english keyboard?

All usage of key here is guarded by `key < 128`

PR #1439, opened 2023-12-11
[7_8] fix crashing caused by missing gs binary on windows

jingkaimori:

Fix the first issue in https://gitee.com/XmacsLabs/mogan/issues/I8MZW1 , fix crashing when exporting given document to pdf. But export fails after this patch.

- [x] handle situation of missing ghostscript under default install position


darcy-shen:
The problem here is why export to PDF will call gs. There are no ps images in the welcome doc.

darcy-shen:
> the first issue 

What is the first issue specified in gitee? Do you mean the gs issue?

jingkaimori:
> > the first issue
> 
> What is the first issue specified in gitee? Do you mean the gs issue?

I mean the crashing issue.

jingkaimori:
gs is invoked by `image_to_pdf`

PR #1447, opened 2023-12-13
[71_10] Qt: translate button label for inputs_list_widget

darcy-shen:


TeXmacs/progs/generic/document-edit.scm
```diff
@@ -74,7 +74,8 @@
 (tm-define (init-interactive-env var)
   (:interactive #t)
   (interactive (lambda (s) (set-init-env var s))
-    (list (or (logic-ref env-var-description% var) var) "string"
+    (list (or (logic-ref env-var-description% var) var)
```
jingkaimori, line 77: Why there are two change of two blank character? These change should be moved to separate pr.

darcy-shen: it is the related code path

darcy-shen: it illustrates how the developer find the bug, and it happened that there are better code formatting in the code path

PR #1462, opened 2023-12-15
[11_15] Unit test to confirm bold and small-caps sys-chinese font resolves ok at first attempt

darcy-shen:


xmake/packages.lua
```diff
@@ -71,13 +88,7 @@ function add_requires_of_mogan()
...
         add_requires("libiconv "..LIBICONV_VERSION, {system=false})
         add_requires("freetype "..FREETYPE_VERSION, {system=false})
-    end
-
-    if is_plat("mingw") or is_plat("windows") then
-        add_requires("qt6widgets "..QT6_VERSION)
-        if is_mode("release") then
-            add_requires("qtifw "..QTIFW_VERSION)
-        end
+        add_requireconfs("pdfhummus.freetype", {version = FREETYPE_VERSION, system = false, override=true})
```
jingkaimori, line 91: why this line is not grouped with other require configs of `pdfhummus`?

jingkaimori (commented):
Generally no problem.

jingkaimori:
maybe no chinese font is installed on archlinux image

xmake/packages.lua
darcy-shen: it works, nevermind

And because there are other customizations per package.

darcy-shen: see https://github.com/XmacsLabs/mogan/pull/1465

PR #1472, opened 2023-12-16
[11_18] Deprecate new font styles option

darcy-shen:


jingkaimori:
`TeXmacs\progs\fonts\font-old-menu.scm` can be removed also.

darcy-shen:
> TeXmacs\progs\fonts\font-old-menu.scm can be removed also.

Can be removed in later pull requests. Not exactly the same thing.

PR #1475, opened 2023-12-16
[71_25] Improve start-up command line message

darcy-shen:


src/System/Config/preferences.cpp
```diff
@@ -118,8 +118,9 @@ load_user_preferences () {
   url prefs_file= "$TEXMACS_HOME_PATH/system/preferences.scm";
   string s;
```
jingkaimori, line 119: Can this variable be removed?

src/System/Misc/data_cache.cpp
```diff
@@ -170,8 +170,8 @@ cache_init (string buffer) {
   if (buffer == "font_cache.scm") {
     url cache_file = url ("$TEXMACS_PATH/fonts/font_cache.scm");
     string cached;
```
jingkaimori, line 172: same as above

src/System/Misc/data_cache.cpp
```diff
@@ -184,8 +184,8 @@ cache_load (string buffer) {
     url cache_file= get_tm_cache_path () * url (buffer);
     //cout << "cache_file "<< cache_file << LF;
     string cached;
```
jingkaimori, line 186: same as above

jingkaimori (commented):
some further simplification can be applied

PR #1525, opened 2023-12-29
[68_8] extract keywords from abstract-data and add them to exported pdf

jingkaimori:

- [x] add a new kind of metadata `keyword`, and corresponding searcher.
- [x] export this metadata to PDF document.
- [ ] related tests for all metadata entries, maybe in separate task

## How to test?

Insert abstraction into a document, and add keywords with focus toolbar. Then export this document to PDF, open exported PDF in browser, and open "document properties" dialogue.

### DIFF
--- a/src/Data/Convert/Texmacs/fromtm.cpp
+++ b/src/Data/Convert/Texmacs/fromtm.cpp
@@ -579,32 +579,41 @@ search_tag (tree t, string tag) {
 }
 
 static string
-search_metadata_tag (tree doc, string tag) {
-  string r;
-  tree t= search_tag (doc, tag);
-  for (int i=0; i<N(t); i++) {
-    if (N(r) != 0) r << ", ";
-    r << tree_to_verbatim (t[i], false, "cork");
+search_metadata_tag_under (tree doc, string parent, string tag) {
+  tree doc_data= search_tag_quick (doc, parent);
+  if (is_compound (doc_data)) {
+    string r;
+    tree   t= search_tag (doc, tag);
+    for (int i= 0; i < N (t); i++) {
+      if (N (r) != 0) r << ", ";
+      r << tree_to_verbatim (t[i], false, "cork");
+    }
+    return r;
   }
-  return r;
+  else {
+    return "";
+  }
 }
 
 string
 search_metadata (tree doc, string kind) {
-  tree doc_data= search_tag_quick (doc, "doc-data");
-  if (doc_data != "") {
-    if (kind == "title")
-      return search_metadata_tag (doc_data, "doc-title");
-    if (kind == "author")
-      return search_metadata_tag (doc_data, "author-name");
-  }
   if (kind == "title") {
-    tree t= search_tag_quick (doc, "tmdoc-title");
-    if (t != "") return tree_to_verbatim (t[0], false, "cork");
+    string r= search_metadata_tag_under (doc, "doc-data", "doc-title");
+    if (is_empty (r)) {
+      tree t= search_tag_quick (doc, "tmdoc-title");
+      if (t != "") return tree_to_verbatim (t[0], false, "cork");
+      else return "";
+    }
+    else {
+      return r;
+    }
   }
-  tree abstract_data= search_tag_quick (doc, "abstract-data");
-  if (!is_empty (abstract_data)) {
-    if (kind == "keyword") {
+  else if (kind == "author") {
+    return search_metadata_tag_under (doc, "doc-data", "author-name");
+  }
+  else if (kind == "keyword") {
+    tree abstract_data= search_tag_quick (doc, "abstract-data");
+    if (is_compound (abstract_data)) {
       for (int i= 0; i < N (abstract_data); i++) {
         if (is_compound (abstract_data[i], "abstract-keywords")) {
           tree abstract_keywords= abstract_data[i];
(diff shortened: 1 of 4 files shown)


src/Data/Convert/Texmacs/fromtm.cpp
```diff
@@ -602,5 +602,25 @@ search_metadata (tree doc, string kind) {
     tree t= search_tag_quick (doc, "tmdoc-title");
     if (t != "") return tree_to_verbatim (t[0], false, "cork");
   }
+  tree ad= search_tag_quick (doc, "abstract-data");
```
darcy-shen, line 605: `ad` is not a good variable name, just use `abstract_data`

src/Data/Convert/Texmacs/fromtm.cpp
```diff
@@ -602,5 +602,25 @@ search_metadata (tree doc, string kind) {
...
+  if (ad != "") {
```
darcy-shen, line 606: `!is_empty (abstract_data)` would be better

src/Data/Convert/Texmacs/fromtm.cpp
```diff
@@ -602,5 +602,25 @@ search_metadata (tree doc, string kind) {
...
+            string r= tree_to_verbatim (ak[0], false, "cork");
+            for (int i= 1; i < N (ak); i++) {
+              r << ", " << tree_to_verbatim (ak[i], false, "cork");
+            }
+            return r;
+          }
+        }
+      }
+    }
+  }
```
darcy-shen, line 624: unit tests should be added for search_metadata

src/Data/Convert/Texmacs/fromtm.cpp
darcy-shen: `dd` is also not a good variable name, should use `doc_data` instead

jingkaimori: what about `ak`?

src/Data/Convert/Texmacs/fromtm.cpp
```diff
@@ -591,16 +591,36 @@ search_metadata_tag (tree doc, string tag) {
 
 string
 search_metadata (tree doc, string kind) {
-  tree dd= search_tag_quick (doc, "doc-data");
-  if (dd != "") {
+  tree doc_data= search_tag_quick (doc, "doc-data");
+  if (doc_data != "") {
```
darcy-shen, line 595: doc_data is a tree, `doc_data != ""` is not a good practice. There are implicit conversion. I guess `N(doc_data) > 0` is better

src/Data/Convert/Texmacs/fromtm.cpp
```diff
@@ -591,16 +591,36 @@ search_metadata_tag (tree doc, string tag) {
...
     if (kind == "author")
-      return search_metadata_tag (dd, "author-name");
+      return search_metadata_tag (doc_data, "author-name");
   }
   if (kind == "title") {
     tree t= search_tag_quick (doc, "tmdoc-title");
     if (t != "") return tree_to_verbatim (t[0], false, "cork");
   }
+  tree abstract_data= search_tag_quick (doc, "abstract-data");
+  if (!is_empty (abstract_data)) {
```
darcy-shen, line 606: same as doc_data

darcy-shen: I thought abstract_data is string, `is_empty` is for string and I remember it is not for tree

`is_empty` is identical to `N(xxx) == 0`, and I think it is better than `N(xxx) == 0`, for tree, we may need to add the is_empty inline routine.

src/Data/Convert/Texmacs/fromtm.cpp
```diff
@@ -591,16 +591,36 @@ search_metadata_tag (tree doc, string tag) {
 
 string
 search_metadata (tree doc, string kind) {
-  tree dd= search_tag_quick (doc, "doc-data");
```
darcy-shen, line 594: the branching in this routine is a bit messed up

the main condition should be kind:

if (kind == "title") {
} else if (kind == "author") {
} else if (kind == "keyword") {
} else {
  error handling on invalid value
}

src/Data/Convert/Texmacs/fromtm.cpp
jingkaimori: `search_tag_quick` may return atomic tree with empty string, on which `N(t)` procedure will crash. `is_compound(doc_data)` may be better choice.

tests/Data/Convert/convert_test.cpp
```diff
@@ -0,0 +1,69 @@
+/******************************************************************************
+ * MODULE     : converter_test.cpp
+ * DESCRIPTION: Properties of characters and strings
+ * COPYRIGHT  : (C) 2019 Darcy Shen
```
darcy-shen, line 4: should be your own name

tests/Data/Convert/convert_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+                                   "keyword 2", "keyword 3")))
+      << string ("Test of Metadata") << string ("author1, author2, author3")
+      << string ("keyword 1, keyword 2, keyword 3");
+  QTest::newRow ("texmacs document")
+      << tree (DOCUMENT, compound ("tmdoc-title", "Test of manual title"))
+      << string ("Test of manual title") << string ("") << string ("");
+}
+
+void
+TestConverter::test_search_metadata () {
```
darcy-shen, line 58: what if the kind is invalid

src/Data/Convert/Texmacs/fromtm.cpp
darcy-shen: just use a long name

PR #1731, opened 2024-03-15
[29_9] Fix for a couple of bugs: 

mgubi:
## What
- always write catalog (which was reason of a bug with newer version of the PDF output) 
- take care of embedded images

## Why
- the Catalog of the modified PDF was not written in all occasions, this was due to wrong use of the interface of the PDF Writer extenders
- images embedded in the document should be ignored when gathering the external attachments paths

## How to test your changes?
- export/import a file with PDF version > 1.4 and with embedded images

### DIFF
--- a/src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
+++ b/src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
@@ -32,7 +32,6 @@
 using namespace PDFHummus;
 using namespace IOBasicTypes;
 
-
 bool
 extract_attachments_from_pdf (url pdf_path, list<url>& names) {
   EStatusCode status= PDFHummus::eSuccess;
@@ -50,28 +49,34 @@ extract_attachments_from_pdf (url pdf_path, list<url>& names) {
         parser.QueryDictionaryObject (parser.GetTrailer (), "Root"));
     // return 0;
     if (!catalog) {
-      if (DEBUG_CONVERT) debug_convert << "Can't find catalog. fail\n";
+      if (DEBUG_CONVERT) debug_convert << "Can't find catalog. fail" << LF;
       status= PDFHummus::eFailure;
       break;
     }
 
     PDFObjectCastPtr<PDFDictionary> d_1 (catalog->QueryDirectObject ("Names"));
     if (!d_1) {
-      if (DEBUG_CONVERT) debug_convert << "Can't find Names dictionary. fail\n";
+      if (DEBUG_CONVERT)
+        debug_convert << "Can't find Names dictionary. fail" << LF;
       status= PDFHummus::eFailure;
       break;
     }
 
-    PDFObjectCastPtr<PDFDictionary> d_2 (d_1->QueryDirectObject ("EmbeddedFiles"));
+    PDFObjectCastPtr<PDFDictionary> d_2 (
+        d_1->QueryDirectObject ("EmbeddedFiles"));
     if (!d_2) {
-      if (DEBUG_CONVERT) debug_convert << "Can't find /Names/EmbeddedFiles dictionary. fail\n";
+      if (DEBUG_CONVERT)
+        debug_convert << "Can't find /Names/EmbeddedFiles dictionary. fail"
+                      << LF;
       status= PDFHummus::eFailure;
       break;
     }
 
     PDFObjectCastPtr<PDFArray> arr (d_2->QueryDirectObject ("Names"));
     if (!arr) {
-      if (DEBUG_CONVERT) debug_convert << "Can't find /Names/EmbeddedFiles/Names array. fail\n";
+      if (DEBUG_CONVERT)
+        debug_convert << "Can't find /Names/EmbeddedFiles/Names array. fail"
+                      << LF;
       status= PDFHummus::eFailure;
       break;
     }
@@ -79,12 +84,12 @@ extract_attachments_from_pdf (url pdf_path, list<url>& names) {
     unsigned long n= arr->GetLength ();
     // Every two elements in the array represent an attachment
     if (n == 0) {
-      if (DEBUG_CONVERT) debug_convert << "arr->GetLength () is 0\n";
+      if (DEBUG_CONVERT) debug_convert << "arr->GetLength () is 0" << LF;
       status= PDFHummus::eFailure;
       break;
     }
     if (n & 1) {
-      if (DEBUG_CONVERT) debug_convert << "arr->GetLength () is wrong\n";
+      if (DEBUG_CONVERT) debug_convert << "arr->GetLength () is wrong" << LF;
       status= PDFHummus::eFailure;
       break;
     }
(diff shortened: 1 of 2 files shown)


src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
```diff
@@ -214,10 +217,10 @@ get_url_image_or_include_tree (tree t, url path) {
     return pre_url;
   }
   else {
-    if (DEBUG_CONVERT)
-      debug_convert << t << " image or include tree format wrong\n" << LF;
-    return url ();
+    if (is_func(t, moebius::INCLUDE) && (DEBUG_CONVERT))
```
darcy-shen, line 220: nit: use `DEBUG_CONVERT` first and then test if it is include

src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
```diff
@@ -214,10 +217,10 @@ get_url_image_or_include_tree (tree t, url path) {
...
+      debug_convert << t << " include tree format wrong\n" << LF;
```
darcy-shen, line 221: nit: use LF instead of `\n`

src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
```diff
@@ -267,7 +270,8 @@ get_linked_file_paths (tree t, url path) {
   array<url> tm_and_linked_file;
   string     label= get_label (t);
   if (label == "image" || label == "include") {
-    tm_and_linked_file << get_url_image_or_include_tree (t, path);
+    url incl_url= get_url_image_or_include_tree (t, path);
+    if (incl_url != url ()) tm_and_linked_file << incl_url;
```
darcy-shen, line 274: if we return `url_none` in `get_url_image_or_include_tree`, we can use `is_none` here

src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
```diff
@@ -305,8 +309,8 @@ replace_url_image_or_include_tree (tree t, url path) {
     t[0]->label= string (name);
   }
   else {
-    if (DEBUG_CONVERT)
-      debug_convert << t << " image or include tree format wrong\n" << LF;
+    if (is_func(t, moebius::INCLUDE) && (DEBUG_CONVERT))
```
darcy-shen, line 312: nit: `DEBUG_CONVERT && is_func () ` would be better

src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
```diff
@@ -305,8 +309,8 @@ replace_url_image_or_include_tree (tree t, url path) {
...
+      debug_convert << t << " include tree format wrong\n" << LF;
```
darcy-shen, line 313: nit: using LF instead of `\n`

darcy-shen:
Use `bin/format` to pass the Lint CI, `clang-format` 16.x is prefered. If you didn't install elvish and do not want to use elvish, you may invoke `clang-format -i src/**` for a in-place formatting.

src/Plugins/Pdf/pdf_hummus_make_attachment.cpp
```diff
@@ -197,6 +207,9 @@ PDFAttachmentWriter::OnCatalogWrite (
     DictionaryContext*  inCatalogDictionaryContext,
     ObjectsContext*     inPDFWriterObjectContext,
     DocumentContext*    inPDFWriterDocumentContext) {
+
+  if (DEBUG_CONVERT) debug_convert << "Populating the EmbeddedFiles dictionary\n";
```
darcy-shen, line 211: nit: using LF instead of `\n`

src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
```diff
@@ -55,24 +57,25 @@ extract_attachments_from_pdf (url pdf_path, list<url>& names) {
...
       status= PDFHummus::eFailure;
       break;
     }
-    PDFObjectCastPtr<PDFDictionary> d_2 (
-        d_1->QueryDirectObject ("EmbeddedFiles"));
+
+    PDFObjectCastPtr<PDFDictionary> d_2 (d_1->QueryDirectObject ("EmbeddedFiles"));
     if (!d_2) {
-      if (DEBUG_CONVERT) debug_convert << "Can't find d2. fail\n";
+      if (DEBUG_CONVERT) debug_convert << "Can't find /Names/EmbeddedFiles dictionary. fail\n";
```
darcy-shen, line 67: nit: use LF

src/Plugins/Pdf/pdf_hummus_make_attachment.cpp
```diff
@@ -80,17 +84,17 @@ pdf_hummus_make_attachments (url pdf_path, array<url> attachment_paths,
   }
 
   PDFWriter   pdfWriter;
+  PDFAttachmentWriter attachmentWriter (&pdfWriter);
   EStatusCode status;
   do {
     status= pdfWriter.ModifyPDF (as_charp (as_string (pdf_path)), ePDFVersion16,
-                                 as_charp (as_string (out_path)));
+                                 as_charp (as_string (out_path)), LogConfiguration(true, false, "PDFWriterLog.txt"));
```
darcy-shen, line 91: Where will `PDFWriterLog.txt` be logged? Could we turn it on only when DEBUG_CONVERTER is enabled?

src/Plugins/Pdf/pdf_hummus_make_attachment.cpp
```diff
@@ -179,6 +182,13 @@ PDFAttachmentWriter::PDFAttachmentWriter (PDFWriter* inPDFWriter) {
   ListenOnCatalogWrite ();
 }
 
+bool 
+PDFAttachmentWriter::IsCatalogUpdateRequiredForModifiedFile (PDFParser* inModifiderFileParser) {
```
darcy-shen, line 186: this is the key to fix the bug!

src/Plugins/Pdf/pdf_hummus_make_attachment.cpp
mgubi: sorry, I should remove it. It was a debugging aid. I guess we need a more well designed mechanism to log the PDFWriter debug trace.

src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
```diff
@@ -267,7 +276,8 @@ get_linked_file_paths (tree t, url path) {
   array<url> tm_and_linked_file;
   string     label= get_label (t);
   if (label == "image" || label == "include") {
-    tm_and_linked_file << get_url_image_or_include_tree (t, path);
+    url incl_url= get_url_image_or_include_tree (t, path);
+    if (incl_url != url ()) tm_and_linked_file << incl_url;
```
darcy-shen, line 280: use `!is_none (incl_url)` here?

darcy-shen (approved):
LGTM except a predicate on url

src/Plugins/Pdf/pdf_hummus_extract_attachment.cpp
darcy-shen: just checked, `url()` do equal to `url_none`, and I need to add a unit test in lolly to make sure the equality works.

darcy-shen:
For `url_none ()` and `url ()`, see https://github.com/XmacsLabs/lolly/pull/297

darcy-shen:
Verified the pr by exporting release_1.2.6.tm to PDF with tm attachments when PDF version is 1.7, it works fine.

PR #1754, opened 2024-03-28
[75_22] Upgrade lolly to 1.4.18 and use string_u8 in converter

darcy-shen:
## What
+ Upgrade lolly to 1.4.18
   + as_int now support `+1`, `+10` by @jingkaimori 
+ Use string_u8 in converter

## Why
string_u8 as an alias of string is just a indicator.

## How to test your changes?
No need to test.

### DIFF
--- a/src/Data/String/converter.cpp
+++ b/src/Data/String/converter.cpp
@@ -365,7 +365,7 @@ sourcecode_to_cork (string input) {
   return output;
 }
 
-string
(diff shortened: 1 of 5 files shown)


jingkaimori (approved):
LGTM, further indication can be made on other api such as `process_keypress_event`

Oyyko:
LGTM

PR #1756, opened 2024-03-29
[7_11] Fix typesetting failure caused by `ones (0)` in the Octave session

darcy-shen:
## What
Fix typesetting failure caused by `ones (0)` in the Octave.

This pull request fixed the bug in the Octave side. It does not fix the bug in the typesetter.

## Why
Fixing the bug in the typesetter is much harder. Fix the bug first and then find the best way to fix it.

## How to test your changes?
See `TeXmacs/tests/11_7.tm`.


Oyyko (approved):
LGTM

PR #1757, opened 2024-03-29
[65_10] Qt: deprecate Qt::AA_UseHighDpiPixmaps

darcy-shen:
## What
```
warning: ‘Qt::AA_UseHighDpiPixmaps’ is deprecated: High-DPI pixmaps are always enabled. This attribute no longer has any effect. [-Wdeprecated-declarations]
  185 |   qApp->setAttribute (Qt::AA_UseHighDpiPixmaps);
      |                          ^~~~~~~~~~~~~~~~~~~~
```
## Why
code cleaning

## How to test your changes?
No need.


jingkaimori (approved):
LGTM, this attribute only takes effect on qt5, which is maintained at 1.2.5.x

PR #1758, opened 2024-03-29
[77_12] Default: page-type defaults to 16:9 when switching to beamer style

darcy-shen:
## What
Switch to `16:9` when switching to beamer style for a new TeXmacs doc.

## Why
`16:9` is more widely adopted than `4:3`

## How to test your changes?
+ Create a new document, and switch to beamer style, check the page type in the focus toolbar (should be 16:9)
+ Open `TeXmacs/tests/tm/77_12_4x3.tm`, the page type should be `4:3`
+ Open `TeXmacs/tests/tm/77_12_16x9.tm`, the page type should be `16:9`

### DIFF
--- a/TeXmacs/packages/beamer/presentation.ts
+++ b/TeXmacs/packages/beamer/presentation.ts
@@ -46,7 +46,7 @@
     </src-comment>
   </active*>
 
-  <assign|page-type|4:3>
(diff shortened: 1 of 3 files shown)


Oyyko (approved):
LGTM

PR #1759, opened 2024-03-29
[77_9] Beamer: fit-to-screen-width after open-buffer of beamer tm doc in book page rendering

darcy-shen:
## What
If the buffer is in beamer style and in book page rendering, `fit-to-screen-width` after `open-buffer`.

## Why
Improve user experience of beamer style.

## How to test your changes?
+ Open `TeXmacs/doc/about/mogan/beamer.en.tm`, it will `fit-to-screen-width`.
+ Open `TeXmacs/doc/about/mogan/research.en.tm`, nothing happened.


Oyyko (approved):
good job!

PR #1760, opened 2024-03-29
[77_13] Beamer: fit-to-screen-width after entering presentation-mode

darcy-shen:
## What
`fit-to-screen-width` after clicking `View->Presentation mode`

## Why
Do not need to zoom out. Before it is `fit-to-screen`, we need to zoom out to make it work.

## How to test your changes?
Open `TeXmacs/doc/about/mogan/beamer.en.tm` and click `View->Presentation mode`.
And then, quit the Presentation mode, and switch to the Slide page rendering, and click `View->Presentation mode` again.


Oyyko (approved):
LGTM

PR #1762, opened 2024-03-29
[66_3] Deprecate the tmml format

darcy-shen:
## What
as title

## Why
`tmu` format will be created for unicode support, we do not need the tmml format now.

## How to test your changes?
+ `File->Export->Xml` should be removed
+ `File->Import->Xml` should be removed
+ `File->Open` does not support tmml file now


Oyyko:
> File->Open does not support tmml file now

Will this cause compatibility issues? Like if a user try to open an old tmml file.

Oyyko:
https://github.com/XmacsLabs/mogan/blob/4211b663dbe488442d6a18c4d3bd3b4a2c894500/TeXmacs/plugins/code/progs/tm-convert-test.scm#L53

We can delele this line?

Oyyko:
Is tmml a format just designed to support unicode? Or does it represent XML format. Will the format of tmu be XML?

Oyyko:
We have a TMML wiki also. lol.

darcy-shen:
TMML is a buggy format. It is not implemented by the main TeXmacs developers, and it is not well maintained.

darcy-shen:
> Will this cause compatibility issues? Like if a user try to open an old tmml file.

Yes. In this case, let the user use GNU TeXmacs or Mogan v1.2.5 LTS.

darcy-shen:
> We can delele this line?

Yes. It has been deleted.

darcy-shen:
> We have a TMML wiki also. lol.

TMML wiki will use tmu format. And `tmml.wiki` is better than `tmu.wiki` as a domain name.

jingkaimori:
"Export selection as image" menu does not show dropdown submenu on mogan before and after this pr, which invokes tmimage.scm and tmmlout.scm. Should this behavior be fixed?

darcy-shen:
> "Export selection as image" menu does not show dropdown submenu on mogan before and after this pr, which invokes tmimage.scm and tmmlout.scm. Should this behavior be fixed?

yes but not in this pr

darcy-shen:
2 days past, let me merge it as an admin.

PR #1763, opened 2024-03-29
[38_8] Deprecate the usage of call_imagemagick_convert

darcy-shen:
## What
+ Deprecate `call_imagemagick_convert` when converting image formats to pdf format
+ Deprecate `call_imagemagick_convert` when exporting the selected document

## Why
Unify the image format conversion to image plugin.

## How to test your changes?
Make sure `(has-binary-convert?)` is true and then:
+ Export `TeXmacs/tests/tm/38_8.tm` to PDF
+ Export the selected part in `TeXmacs/tests/tm/38_8.tm` as the supported image format


Oyyko (approved):
LGTM

PR #1764, opened 2024-03-29
[78_4] Fix non-void function does not return a value in all control p…

darcy-shen:
…aths in QTMPixmapOrImage


darcy-shen:
It is a trivial pull request, that's why I merge it without review.

PR #1765, opened 2024-03-30
[67_2] Octave: the binary plugin

darcy-shen:

PR #1772, opened 2024-04-01
[75_12] Fix url_unix since lolly 1.4.0

darcy-shen:
## What
Fix the glue `url-unix`.

## Why
Before of the semantic change in lolly 1.4.0. The default constructor of url means `system->url` now.

## How to test your changes?
+ [x] Windows
+ [ ] macOS
+ [ ] Linux
Test if the maxima or octave plugin works.

### DIFF
--- a/src/Scheme/L2/glue_lolly.lua
+++ b/src/Scheme/L2/glue_lolly.lua
@@ -507,7 +507,7 @@ function main()
             },
             {
                 scm_name = "url-unix",
-                cpp_name = "url",
+                cpp_name = "url_unix",
                 ret_type = "url",
                 arg_list = {
                     "string",


darcy-shen:
This is a hot fix, it blocks #1765 and #1766 , let me merge it now!

jingkaimori:
After this fix, "Export selection as image" menu shows dropdown submenu, but `svg` is not appeared in list. If suffix of exported image is changed to `svg`, mogan will prompt that "Sorry, pdf to svg converter is missing. Generating pdf instead", and `serialize-tmml` function will not be called

darcy-shen:
Yes. Without this pull request, the plugin system on Windows is broken on v1.2.6

darcy-shen:
> but svg is not appeared in list. 

It means, no known converter found from pdf to svg. I doubt if svg exporting should be supported. The current impl relies on the PDF to SVG converter.

PR #1780, opened 2024-04-02
[60_8] Maxima: remove snap installed one from binary candidates

darcy-shen:
## What
as title

## Why
`/snap/bin/maxima -p /usr/share/xxx.lisp` does not work 

## How to test your changes?
Build and install deb, try to launch maxima the via the snap installed one.


Oyyko (approved):
lgtm

PR #1784, opened 2024-04-04
[65_11] Qt 6: deprecate QVariant::canConvert(int targetTypeId)

darcy-shen:
## What
https://doc.qt.io/qt-6/qvariant-obsolete.html#canConvert-1
> `bool QVariant::canConvert(int targetTypeId) const`
> This function is deprecated since 6.0. We strongly advise against using it in new code.
> This is an overloaded function.
> Use canConvert(QMetaType(targetTypeId)) instead.

## How to test your changes?
CI

### DIFF
--- a/src/Plugins/Qt/QTMMenuHelper.cpp
+++ b/src/Plugins/Qt/QTMMenuHelper.cpp
@@ -27,6 +27,7 @@
 #include <QApplication>
 #include <QCompleter>
 #include <QKeyEvent>
+#include <QMetaType>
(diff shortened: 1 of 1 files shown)


jingkaimori (approved):
LGTM, but maybe `d.canConvert<QString> ()` is better here?

darcy-shen:
> but maybe d.canConvert<QString> () is better here?

I don't know, it works, and it eliminates the build warnings, that's all I care.

PR #1786, opened 2024-04-04
[29_9] Use is_none instead of u != url ()

darcy-shen:
## What
Use `is_none` instead of `u != url ()`

## Why
`u != url ()` is not a good way to express `!is_none`, `url () == url_none ()` is a hidden rule of the url class.

## How to test your changes?
CI


jingkaimori (approved):
LGTM

PR #1787, opened 2024-04-04
[66_6] Protocol: scheme_u8 for UTF-8 Scheme tree

darcy-shen:
## What
This is a protocol change, it brings new feature which is absent for GNU TeXmacs <= 2.1.4.

Add `scheme_u8` input mode for UTF-8 Scheme tree.

## Why
Use UTF-8 Scheme tree as the protocol for plugins.

## How to test your changes?
See `TeXmacs/tests/tm/66_6.tm`

How it works:
1. `66_6.tm` is encoded in cork
2. `python-serialize` in `init-python.scm` will convert the cork encoded tree into UTF-8 python code snippet
3. The python interpreter evaluates the UTF-8 python code snippet and generate the UTF-8 output
4. The output is captured by the "fragile" protocol defined in `input.cpp`

`tree_utf8_to_cork` is an adhoc impl, it will be migrated to moebius later. Please focus on the protocol design to review this pr. In scala snippet, `tree_utf8_to_cork` should be implemented as `tree.map (utf8_to_cork)`. The C++'s syntax is too ugly for functional programming. I don't know if it is possible for us to add several sweet functional interfaces in lolly.

### DIFF
--- a/TeXmacs/tests/tm/66_6.tm
+++ b/TeXmacs/tests/tm/66_6.tm
@@ -43,7 +43,7 @@
     <|unfolded-io>
       flush_any ("scheme_u8:(frac \<#5206\>\<#5B50\> \<#5206\>\<#6BCD\>)")
     <|unfolded-io>
-      <frac|分子|分母>
+      <frac|\<#5206\>\<#5B50\>|\<#5206\>\<#6BCD\>>
     </unfolded-io>
 
     <\input>
--- a/TeXmacs/tests/tm/66_6.tm
+++ b/TeXmacs/tests/tm/66_6.tm
@@ -1,0 +1,64 @@
+<TeXmacs|2.1.2>
+
+<style|<tuple|generic|no-page-numbers|chinese|python>>
+
+<\body>
+  <\session|scheme|default>
+    <\unfolded-io|Scheme] >
+      (stree-\<gtr\>tree `(frac 1 2))
+    <|unfolded-io>
+      <text|<frac|1|2>>
+    </unfolded-io>
+
+    <\input|Scheme] >
+      \;
+    </input>
+  </session>
+
+  <\session|python|default>
+    <\output>
+      Python 3.12.2 [/usr/bin/python3]
+
+      Python plugin for TeXmacs.
+
+      Please see the documentation in Help -\<gtr\> Plugins -\<gtr\> Python
+    </output>
+
+    <\input>
+      \<gtr\>\<gtr\>\<gtr\>\ 
+    <|input>
+      from tmpy.protocol import flush_any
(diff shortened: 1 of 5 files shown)


Oyyko:
What is `Add scheme_u8 input mode for UTF-8 Scheme tree.`

I thought we were going to support reading in files encoded in UTF8. `66_6.tm` is still cork encoded. Sorry, maybe I didn’t understand the content of this PR.

darcy-shen:
> What will happen after The output is captured by the "fragile" protocol defined in input.cpp. Where will the output be written?

Currently, Scheme session is not running in a separated process. That's why when I'm teaching the interactive SICP, it will freeze Mogan if we trigger an infinite loop. And Python session is running in a separated process. For Inter-process communication, we are using Pipe in TeXmacs and the related code is not wrapped in sdk. That's why it is fragile.

In reality, for IPC, RPC is much more robust. For example, in Apache Zeppelin, Thrift is adopted. And in Apache Spark, netty is adopted. gRPC seems to be a more popular option because its nice implementations in major programming languages.

In GNU TeXmacs, here is the current impl:
https://github.com/XmacsLabs/mogan/blob/branch-1.2/src/Plugins/Qt/qt_pipe_link.cpp

Ideally, plugins should work fine on a remote server. I've never tried remote plugins of GNU TeXmacs.

When talking about plugin system, Apache Zeppelin is similar to GNU TeXmacs. The data exchange between the plugin and the host is not expected to be big.

darcy-shen:
> What is Add scheme_u8 input mode for UTF-8 Scheme tree.
> I thought we were going to support reading in files encoded in UTF8. 66_6.tm is still cork encoded. Sorry, maybe I didn’t understand the content of this PR.

`stree` data is the `json` equiv in GNU TeXmacs. For a historical reason, the label whose data type is string and is only available on the leaf of the tree is encoded in cork encoding.

`stm` doc format is the format which is serialized from cork string label. We are going to add a `smg` format which is serialized from `(cork->string the_label)`.

And this pull request unlocks the UTF-8 scheme protocol for plugins. Before, the scheme input mode is encoded in cork encoding. See https://github.com/XmacsLabs/mogan/pull/1756

Oyyko (approved):
LGTM

darcy-shen:
For the [TMU format](https://github.com/XmacsLabs/tmu), when SMG format is ready, we could design a data-loss free way to convert TMU to SMG and convert SMG back to TMU. SMG format is the most trust-worthy data format for Mogan. And TMU format is not that trust-worthy, because it is design for human beings and version control system. When we try to make both human beings and VCS happy, complexity is increased inevitablely.

PR #1792, opened 2024-04-07
[61_3] Binary: version-binary-xyz impl

darcy-shen:
## What
+ Common routines `(version-binary)`

Return the first non-empty line as the version line for the binary.

## Why
Binary version is needed for all binaries.

## How to test your changes?
Open `TeXmacs/tests/tm/61_3.tm`.


Oyyko (approved):
LGTM

PR #1798, opened 2024-04-09
[61_3] Binary: candidates-only and off option for a binary

darcy-shen:
## What
Added two options of `plugin:binary:xyz`:
+ `candidates-only`: only resolve candidates, if not found, just return `(url-none)`
+ `off`: disable the binary plugin, just return `(url-none)`

## Why
Both of the options are for the security reason:
+ `candidates-only`: trust the candidates
+ `off`: do not trust the binary

## How to test your changes?
See `TeXmacs/tests/tm/61_3.scm`.


Oyyko (approved):
LGTM

PR #1744, opened 2024-03-22
[61_3] Binary plugins: find-binary

darcy-shen:
## What
+ impl common routine `find-binary` in the following order:
   1. Check `plugin:binary:<binary-name>` first
   2. Check `(binary-gs-candidates)` secondly
   3. Check binary under the PATH (excluding im-proper PATH like `C:/Windows/System32`)
+ Migrate all binary plugins to `(find-binary)`
+ Added two path candidates for ImageMagick convert on Windows

## Why
+ Fix convert bug on Windows
+ Reduce duplicated code

## How to test your changes?
+ [x] macOS
+ [x] Windows
+ [ ] Linux

Open `TeXmacs/tests/tm/61_3.tm` and evaluate the Scheme sessions.

PR #1800, opened 2024-04-09
[61_3] Binary: improve impl for finding binary in sys path

da-liii:
## What
Re-impl finding binary in sys path.

Related pull request: https://github.com/XmacsLabs/mogan/pull/1744

## Why
We must keep the consistency of `binary-id`. In the previous impl, the binary-id for python3 is `python3` on linux and macOS, and is `python` on windows.

## How to test your changes?
Open `TeXmacs/tests/tm/61_3.tm`


Oyyko (approved):
LGTM

PR #1803, opened 2024-04-12
[79_2] IM: disable preedit by default on linux to speedup editing

da-liii:
## What
Disable IM preedit on Linux to speedup.

Menu entry to turn on/turn of IM pre-editing will be implemented in another pull request.

## Why
Performance tuning.

Why not turn it on by default on macOS and Windows: because input via speech is available on Windows and macOS.

## How to test your changes?
+ [x] Linux
+ [ ] macOS
+ [ ] Windows


Oyyko (approved):
LGTM

PR #1804, opened 2024-04-13
[78_2] Revert: Gitee Scan Rule 025: Effective C++, item 30: understan…

da-liii:
…d the ins and outs of inlining

## What
+ Revert the removal of inline for `inline virtual xxx()`.
+ Revert the removal of virtual for `inline virtual yyy()`.

## Why
The removal of `virtual` in `inline virtual derived_this ()` leads to a segment fault.

## How to test your changes?
+ [x] Linux
+ [ ] macOS
+ [ ] Windows

Without this pull request:

1. Open `Help->Plugins->Binary`
2. Use the Replace widget to replace `python3` with `gs`
3. Close the replace widget
4. Open `Help->Plugins->Maxima` and then there will be a segmentation fault at `src/Edit/Interface/edit_repaint.cpp:430`

With this pull request, it will be fixed.

### DIFF
--- a/src/Edit/Editor/edit_main.hpp
+++ b/src/Edit/Editor/edit_main.hpp
@@ -43,7 +43,7 @@ private:
 public:
   edit_main_rep (server_rep* sv, tm_buffer buf);
   ~edit_main_rep ();
-  inline void* derived_this () { return (edit_main_rep*) this; }
+  virtual inline void* derived_this () { return (edit_main_rep*) this; }
 
   void        set_property (scheme_tree what, scheme_tree val);
   void        set_bool_property (string what, bool val);
--- a/src/Graphics/Gui/widget.hpp
+++ b/src/Graphics/Gui/widget.hpp
@@ -38,8 +38,8 @@ protected:
 public:
   widget_rep ();
   virtual ~widget_rep ();
-  inline void*        derived_this () { return (widget_rep*) this; }
-  virtual tm_ostream& print (tm_ostream& out);
+  inline virtual void* derived_this () { return (widget_rep*) this; }
+  virtual tm_ostream&  print (tm_ostream& out);
 
   virtual void send (slot s, blackbox val);
   // send a message val to the slot s
--- a/src/Edit/editor.hpp
+++ b/src/Edit/editor.hpp
@@ -136,7 +136,7 @@ protected:
 public:
   editor_rep ();
   editor_rep (server_rep* sv, tm_buffer buf);
-  virtual ~editor_rep () {}
+  inline virtual ~editor_rep () {}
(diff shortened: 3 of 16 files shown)


src/Edit/Editor/edit_main.hpp
```diff
@@ -43,7 +43,7 @@ class edit_main_rep : public edit_interface_rep,
 public:
   edit_main_rep (server_rep* sv, tm_buffer buf);
   ~edit_main_rep ();
-  inline void* derived_this () { return (edit_main_rep*) this; }
+  virtual inline void* derived_this () { return (edit_main_rep*) this; }
```
Oyyko, line 46: The code here removed virtual before, not inline?

src/Graphics/Gui/widget.hpp
```diff
@@ -38,8 +38,8 @@ class widget_rep : public abstract_struct {
 public:
   widget_rep ();
   virtual ~widget_rep ();
-  inline void*        derived_this () { return (widget_rep*) this; }
-  virtual tm_ostream& print (tm_ostream& out);
+  inline virtual void* derived_this () { return (widget_rep*) this; }
```
Oyyko, line 41: The code here removed virtual before, too. not removed inline.

Oyyko:
I suspect that the problem is not caused by removing inline in `inline virtual`, but by removing virtual.

Oyyko:
the Replace widget on my computer cannot replace `python3` in session to `gs`. It can only replace the `pythone3` in normal text.

Oyyko:
> the Replace widget on my computer cannot replace `python3` in session to `gs`. It can only replace the `pythone3` in normal text.

Somehow it works. But I don't have the same error on mac.

Oyyko:
In the [78_2] commit, we actually removed `virtual` in `virtual inline`, not inline.

See http://git.tmml.wiki/XmacsLabs/mogan/commit/867dc9aabd1b48679650687f0b8a9e60db2ef630?style=split&whitespace=ignore-change&show-outdated=.

image

src/Edit/Editor/edit_main.hpp
da-liii: yes, let me update the pr desc

da-liii:
> But I don't have the same error on mac.

I can not reproduce it on macOS 14.3.

da-liii:
> I can not reproduce it on macOS 14.3.

I can reproduce it now. But I need to use the replace widget and close it for multiple times.

da-liii:
> the Replace widget on my computer cannot replace python3 in session to gs. It can only replace the pythone3 in normal text.

You need to use the replace widget but not the footer toolbar.

PR #1815, opened 2024-04-16
[61_4] Style: physics v1.0.0

da-liii:
## What
as title

## Why
Make the physics plugin as the built-in plugin

## How to test your changes?
Open `Help->Plugin->Physics`


da-liii:
Because it is a brand new plugin, let me merge it now.

PR #1839, opened 2024-04-22
[56_5] Cache styles by content digest

da-liii:
## What
+ Fix the search result orders for `search_sub_dirs`
  + Side effect: user installed fonts will override the system fonts (it is a fix on the wrong behavior)
+ Cache all styles by content digests
+ Check `Debug->io` for the debug info of style resolving

Here is the expected order to load a style or package:
+ `$TEXMACS_HOME_PATH/styles`
+ `$TEXMACS_PATH/styles`
+ `$TEXMACS_HOME_PATH/plugins/xxx/styles/`
+ `$TEXMACS_PATH/plugins/xxx/styles`
+ `$TEXMACS_HOME_PATH/packages`
+ `$TEXMACS_PATH/packages`
+ `$TEXMACS_HOME_PATH/plugins/xxx/packages/`
+ `$TEXMACS_PATH/plugins/xxx/packages/`

## Why
Content digests should be adopted for modification on user defined styles/packages.

## How to test your changes?
Try customizing https://github.com/XmacsLabs/physics

### DIFF
--- a/src/Data/Document/new_style.cpp
+++ b/src/Data/Document/new_style.cpp
@@ -62,13 +62,22 @@ hashmap<string, bool>        hidden_packages (false);
 static url
 resolve_local_style (string style_name) {
   string pack= style_name * ".ts";
-  return resolve (url ("$TEXMACS_STYLE_PATH") * pack);
+  url    ret = resolve (url ("$TEXMACS_STYLE_PATH") * pack);
+  if (DEBUG_IO) {
+    debug_io << "Resolved local style: " << style_name << " -> " << ret << LF;
+  }
+  return ret;
 }
 
 static url
 resolve_relative_style (string style_name, url master) {
   string pack= style_name * ".ts";
-  return resolve (expand (head (master) * url_ancestor () * pack));
+  url    ret = resolve (expand (head (master) * url_ancestor () * pack));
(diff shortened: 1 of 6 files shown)


src/Data/Document/new_style.cpp
```diff
@@ -123,7 +158,9 @@ cache_file_name_sub (tree t) {
 
 static string
 cache_file_name (tree t) {
-  return cache_file_name_sub (t) * ".scm";
+  url tmp= url_temp ("txt");
+  save_string (tmp, cache_file_name_sub (t));
```
da-liii, line 162: need a new release of lolly to avoid the sink of the digest

Oyyko (approved):
lgtm.

src/Data/Document/new_style.cpp
jingkaimori: Why uses MD5 here? Use SHA256 inside `cache_file_name_sub` is enough.

jingkaimori (commented):
Some suggestions for this pr.

src/Data/Document/new_style.cpp
da-liii: Long file name may crash the file system

PR #1038, opened 2023-09-17
[23_27] Draw an arc by center and two points

Oyyko:
## Why you open this Pull Request?
Draw an arc by center and two points.

Enter three points. The first is the center of the circle and the second is the starting point of the arc. Draw a straight line through the third point and the center of the circle. The intersection of the straight line and the circle is the end point of the arc. Use clockwise method.

## What work have you done in the current Pull Request?
- [x] Draw an arc by center and two points.

PR #1842, opened 2024-05-01
[23_27] Change std-arc to counterclockwise

Oyyko:
## What
Change the arc direction of std-arc to counterclockwise
## Why
It's more common for math workers.
## How to test your changes?
Use mogan's arc drawing function

### DIFF
--- a/TeXmacs/progs/graphics/graphics-markup.scm
+++ b/TeXmacs/progs/graphics/graphics-markup.scm
@@ -94,7 +94,7 @@
          (vec-c-q (points-sub q c))
          (m  (if (equal? r 0.0)
                 c
-                (if (> (points-cross-product-k vec-c-p vec-c-q) 0)
+                (if (< (points-cross-product-k vec-c-p vec-c-q) 0)
                   (points-add (point-times (point-get-unit (points-sub mid-p-x c)) (- r)) c)
                   (if (= (points-cross-product-k vec-c-p vec-c-q) 0)
                     ;; If cross product == 0, then the angle between vec-c-p and vec-c-q is 0 or 180.


Oyyko:
Maybe useful reference: #1038

Oyyko:
> Here is the test doc, with or without this pull request, the rendering of the arc is different. Please think about compatibility!
> 
> Must we break the compatibility?
> 
> ```
> <TeXmacs|2.1.2>
> 
> <style|<tuple|generic|no-page-numbers|chinese>>
> 
> <\body>
>   <with|gr-mode|<tuple|edit|point>|gr-frame|<tuple|scale|1cm|<tuple|0.5gw|0.5gh>>|gr-geometry|<tuple|geometry|1par|0.6par>|gr-grid|<tuple|empty>|gr-edit-grid-aspect|<tuple|<tuple|axes|none>|<tuple|1|none>|<tuple|10|none>>|gr-edit-grid|<tuple|empty>|gr-arrow-end|\<gtr\>|gr-grid-old|<tuple|cartesian|<point|0|0>|2>|gr-edit-grid-old|<tuple|cartesian|<point|0|0>|1>|<graphics||<with|arrow-end|\<gtr\>|<std-arc|<point|0|0>|<point|0.0|2.0>|<point|2.0|0.0>>>>>
> </body>
> 
> <\initial>
>   <\collection>
>     <associate|page-medium|paper>
>     <associate|page-screen-margin|false>
>   </collection>
> </initial>
> ```

One solution is to introduce a new data type, for example we can name it std2-arc. std-arc is clockwise, and std2-arc is counterclockwise. Then in the new version we can generate std2-arc by default when drawing an arc. This is a solution. I'm wondering if there is a better solution.

Oyyko:
@da-liii I add a new type called `std-arc-counterclockwise`. And we will use this for drawing an arc by default in this PR. The old document with `std-arc` type can also be drawed correctly. So we maintained the compatibility.

TeXmacs/progs/graphics/graphics-markup.scm
```diff
@@ -106,6 +106,30 @@
     `(arc ,p ,m ,x)))
 
 
+(define-graphics (std-arc-counterclockwise C P Q)
```
da-liii, line 109: please abstract the common part for `std-arc-counterclockwise` and `std-arc`

da-liii:
Is there a better and shorter name for `std-arc-counterclockwise` to express the same idea?

Oyyko:
I've abstracted the common part into a helper function.

TeXmacs/progs/graphics/graphics-markup.scm
```diff
@@ -135,6 +145,35 @@
            (line ,p ,c ,x))
           (arc ,p ,m ,x)))))
 
+(define-graphics (sector-counterclockwise C P Q)
```
da-liii, line 148: for sector, the direction is meaningless, I think

da-liii: There is a same problem here, we should abstract the common part for `sector` and `sector-counterclockwise`

Oyyko: I am trying to fix this. Where can I find the cache of the *.ts file now? it's not in `~/Library/Application\ Support/XmacsLabs`. Are they not cached? Because my change to the .ts file seems not work.

da-liii: see https://github.com/XmacsLabs/mogan/blob/branch-1.2/src/System/Misc/tm_sys_utils.cpp#L173-L188

It is cached in `$HOME/Library/Caches/`

da-liii: the direction is meaningless, but we still need to keep the compatibility

Oyyko:
@da-liii I've added sector counterclockwise.

da-liii (approved):
Tested on Linux

PR #1852, opened 2024-05-09
[20_10] Fix Edit->Copy to->Image by setting texmacs->image:format to png

da-liii:
## What
`Edit->Copy to->Image` does not work because `(get-preference "texmacs->image:format")` is `"default"`

## How to test your changes?
+ [x] Linux
+ [ ] macOS
+ [x] Windows

Select something in the buffer, and click `Edit->Copy to->Image` and then paste it somewhere else.


jingkaimori (approved):
LGTM

PR #1858, opened 2024-05-13
[73_34] CI on Ubuntu 24.04

da-liii:


da-liii:
Ubuntu 24.04 is not available for community now. see https://github.com/github/roadmap/issues/958

PR #1918, opened 2024-06-28
[14_12] Added test cases for src/Data/Parser/string_parser.cpp

UnbSky:
## What
Added test cases for src/Data/Parser/string_parser.cpp
## Why
The current string_parser can handle most strings, but it cannot handle the **'** or **"** in the string
## How to test your changes?
```
xmake build string_parser_test
xmake r string_parser_test
```


da-liii (approved):
based on Trust

PR #1919, opened 2024-07-04
[66_8] remove useless code in to_tmu

jingkaimori:
## What

* use `binary_to_hexadecimal` to optimize performance when saving large images.
* remove line truncate mechanism of tmu format

## Why

`binary_to_hexadecimal` is obviously faster. Threshold of line truncator is set to 65535, so that ordinary document will not trigger truncating in tmu, thus remove it can make code clean and improve performance.

## How to test your changes?

As same as other pr about tmu.


da-liii (approved):
LGTM

PR #1920, opened 2024-07-04
[14_13] Add varparallel and nvarparallel

KeShih:
## What
Add `\varparallel` and `\nvarparallel`.
## Why
These symbols are widely used in plane and solid geometry.
## How to test your changes?
Use `\varparallel` and `\nvarparallel` in math mode.
image


da-liii:
image

It is ok in GNU TeXmacs or Mogan.

da-liii (approved):
Awesome!

PR #1932, opened 2024-07-14
[81_1] Add a blank tab page toolbar to the main window (default invisible currently)

Minzihao:
## What
Add a blank tab page toolbar to the main window. Due to the current blank toolbar, it is not displayed by default. You can display it by the main menu "view" -> "Tab bar".
## Why
I have **modified** 11 files and **added** 2 new files. I understand that the recommended practice is to have small code merge requests. Although I have modified many files, most of these modifications are only a few lines of code and are necessary.

For example, the changes in the following 7 files are all within **3 lines**:

- TeXmacs/progs/init-research.scm
- TeXmacs/progs/texmacs/menus/view-menu.scm
- src/Edit/Interface/edit_interface.cpp
- src/Graphics/Gui/widget.cpp
- src/Texmacs/Data/new_window.cpp
- src/Texmacs/Window/tm_frame.cpp
- src/Texmacs/Window/tm_window.cpp

Why do so many files need to be modified even if the changes are small? Because this follows the existing practice of how the main toolbar is implemented, and I need to follow the same process.

Additionally, the two newly added files, 

- TeXmacs/progs/texmacs/menus/tabpage-menu.scm
- src/Plugins/Qt/QTMTabPage.hpp

are actually placeholder files intended to lay the groundwork for subsequent code.

Therefore, excluding these files, the main file I modified is `qt_tm_widget.cpp`.

## How to test your changes?
Build and run, you can display it by the main menu "view" -> "Tab bar".


Oyyko:
Lint/build check failed. https://github.com/XmacsLabs/mogan/actions/runs/9924852710/job/27416623080?pr=1932

Oyyko:
> Lint/build check failed. https://github.com/XmacsLabs/mogan/actions/runs/9924852710/job/27416623080?pr=1932

I believe it's about that you need to use clang format before upload your code.

TeXmacs/progs/texmacs/menus/view-menu.scm
```diff
@@ -134,6 +134,7 @@
   ---
   (when (visible-header?)
         ("Main icon bar" (toggle-visible-icon-bar 0))
+        ("Tab bar" (toggle-visible-icon-bar 4))
```
Oyyko, line 137: Is there a reason why the number 4 is placed after 0 and before 1? Serial number 4 is just a serial number? And does the order of the codes determine the order in which they are displayed?

TeXmacs/progs/texmacs/texmacs/tm-view.scm
```diff
@@ -49,6 +51,7 @@
 (define-preferences
   ("header" "on" notify-header)
   ("main icon bar" "on" notify-icon-bar)
+  ("tab bar" "off" notify-icon-bar)
```
Oyyko, line 54: Why here off?

src/Texmacs/Window/tm_frame.cpp
```diff
@@ -114,7 +114,7 @@ tm_frame_rep::menu_main (string menu) {
 
 void
 tm_frame_rep::menu_icons (int which, string menu) {
-  if ((which < 0) || (which > 3) || (!has_current_view ())) return;
+  if ((which < 0) || (which > 4) || (!has_current_view ())) return;
```
Oyyko, line 117: Can we add a `constexpr` var or MACRO here to record the number. This way we can prevent the next person who writes this code from having to make manual changes.

Minzihao:
> > Lint/build check failed. https://github.com/XmacsLabs/mogan/actions/runs/9924852710/job/27416623080?pr=1932
> 
> I believe it's about that you need to use clang format before upload your code.

ok

TeXmacs/progs/texmacs/texmacs/tm-view.scm
Minzihao: Due to the current blank toolbar, it is not displayed by default.

TeXmacs/progs/texmacs/menus/view-menu.scm
Minzihao: image
Because the Tab bar is displayed between the Main icon bar and the Mode dependent icons, I want the order of options in the menu to be consistent as well. The code here will only affect the positions of options in the menu.

src/Texmacs/Window/tm_frame.cpp
Minzihao: Ok, I will use macro instead.

TeXmacs/progs/texmacs/menus/view-menu.scm
Minzihao: Do you think it is necessary to place the number 4 after 3?

TeXmacs/progs/texmacs/texmacs/tm-view.scm
Oyyko: Get!

TeXmacs/progs/texmacs/menus/view-menu.scm
Oyyko: > Do you think it is necessary to place the number 4 after 3?

Both are OK if it is reasonable. I think the code now is good.

da-liii:
> I believe it's about that you need to use clang format before upload your code.

see https://github.com/XmacsLabs/mogan/blob/branch-1.2/docs/guide/Format.md

src/Texmacs/Window/tm_frame.cpp
Minzihao: Done✔

Minzihao:
> > I believe it's about that you need to use clang format before upload your code.
> 
> see https://github.com/XmacsLabs/mogan/blob/branch-1.2/docs/guide/Format.md

Done✔

da-liii (approved):
LGTM

PR #1941, opened 2024-07-17
[61_7] S7: rework in Scheme

da-liii:
## What
New feature:
img_v3_02ct_4d5b161e-8121-47db-ab94-a0a62f6e2c0g

The other features remains the same with the cpp impl.

## Why
To make it easier to maintain.

## How
+ Open `61_7.tm`
+ Open `Help -> Plugins -> S7 Scheme`


Oyyko:
LGTM

PR #1948, opened 2024-07-21
[43_2] Minor: Fix check count because of execution order

da-liii:
## What
as title

## Why
The `load-buffer` glued scheme routines will result to a side effect on the S7 Scheme engine.

Use `let` to bind the evaluated value first, and then check. In other words, check after the side effect.

## How to test your changes?
```
xmake r 43_2
```
The success count should be 1 instead of 0.


da-liii:
Here is the screenshot of tests:
image


Now the success count is correct!

ATQlove:
LGTM!

ATQlove (approved):
LGTM!

PR #1974, opened 2024-07-31
[24_14] Fix unhandled slots in headless mode

da-liii:
## What
Fix crash when exporting tm/tmu docs to HTML with the `-headless` option

## Why
When exporting TeXmacs/Mogan documents to html and using the `-headless` option, Mogan will crash for several documents in planet. That's why we added a blacklist in https://github.com/XmacsLabs/planet/pull/37/

## How to test your changes?
```
xmake r 24_14
```


ATQlove (approved):
LGTM

PR #1983, opened 2024-08-03
[81_5] Keep tab-title and doc-path of tab pages untranslated

Minzihao:
## What
Keep tab-title and doc-path of tab pages untranslated.
### Before:
The tab title and doc path of file `session.tm` are translated into Chinese which is not expected.
屏幕截图 2024-08-03 202324

### After:
屏幕截图 2024-08-03 202240


da-liii (approved):
LGTM

PR #1990, opened 2024-08-06
[61_5] Gnuplot: initial rework

da-liii:
## What
Reimplement the Gnuplot plugin in Goldfish Scheme.

## Why
1. Some people have trouble installing Python on Windows, they are complaining that they failed to get the gnuplot plugin work on Windows
2. Less deps, because goldfish is built-in, we only need to install gnuplot

## How to test your changes?
Launch the gnuplot session.


da-liii:
@mgubi Now, I've re-impl the gnuplot plugin in Scheme!

Compared with https://github.com/texmacs/texmacs/blob/master/plugins/tmpy/graph/gnuplot.py

Scheme (R7RS) is much nicer choice to implement a plotting plugin.

PR #2028, opened 2024-08-20
[70_12] Revert: Language: add current ip for advance

jingkaimori:
This reverts commit 361f1c14f74895ade3c59f4c1e9c1ece3596121f.

## Why

inverse path get from argument is the same as ones obtained by `obtain_ip`, thus change is unnecessary.


da-liii (approved):
LGTM

PR #2031, opened 2024-08-21
[20_12] fix corrupted ip(inversed path) when tree is exported to image

jingkaimori:
## What

When some snippet is printed, this snippet is inserted into the_et first, then typesetter is invoked with proper reverse path.

## Why

Because a empty path is passed to typesetter, several error messages about improper path is printed when printing snippets. If at least two blocks of code is selected and exported, mogan will crash.

## How to test your changes?

- [ ] select something in any document, then copy them as image
    - [x] code 
    - [ ] image
    - [ ] ordinary text
- [ ] export document with image into other format
    - [ ] html
    - [ ] tex


da-liii (approved):
LGTM

PR #2038, opened 2024-08-26
[24_17] HTML: docx data plugin

ATQlove:
## What
Add docx export option in the GUI toolbar and improve the docx export process.

## How to test your changes?
Choose a TeXmacs file and export it by "File->Export->docx..."
图片
 Here is a tm file:
图片
Then here is exported docx file:
图片


da-liii (approved):
LGTM

PR #2051, opened 2024-08-30
[24_19] Avoid converting CJK to hexadecimal when exporting html

ATQlove:
## What
Avoid converting Chinese to hexadecimal when exporting html.

## Why
When exported tm, tmu format to html before, Chinese was displayed in hexadecimal. After modification, Chinese can be displayed normally in html.

## How to test your changes?
Text in tmu:
图片
Text in html:
图片
Open html in browser:
图片

### DIFF
--- a/src/Data/String/converter.cpp
+++ b/src/Data/String/converter.cpp
@@ -547,7 +547,39 @@ string
 utf8_to_html (string_u8 input) {
   converter conv= load_converter ("UTF-8", "HTML");
   string    s   = apply (conv, input);
-  return utf8_to_hex_entities (s);
+
+  bool should_convert_to_hex= true; // Transcoding is required by default
+
+  // Iterate over the string and check if it contains CJK characters
+  for (int i= 0; i < N (s);) {
+    unsigned int code= decode_from_utf8 (s, i);
+
+    // Determining whether a character is a CJK character
+    if ((code >= 0x4e00 &&
+         code <= 0x9fff) || // commonly used Chinese characters
+        (code >= 0x3400 && code <= 0x4dbf) ||   // CJK extension A
+        (code >= 0x20000 && code <= 0x2a6df) || // CJK extension B
+        (code >= 0x2A700 && code <= 0x2B73F) || // CJK extension C
+        (code >= 0x2B740 && code <= 0x2B81F) || // CJK extension D
+        (code >= 0x2B820 && code <= 0x2CEAF) || // CJK extension E
+        (code >= 0x2CEB0 && code <= 0x2EBEF) || // CJK extension F
+        (code >= 0xF900 && code <= 0xFAFF) ||   // CJK-compatible characters
+        (code >= 0xAC00 && code <= 0xD7AF) ||   // Korean alphabet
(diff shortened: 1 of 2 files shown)


da-liii:
这个pr有问题，不应该在这个函数里面改

ATQlove:
Modified for html only, and expanded normal display characters in html to cjk.

Text in tmu:
图片
Text in html:
图片
Open html in browser:
图片

src/Data/String/converter.cpp
```diff
@@ -547,7 +547,39 @@ string
...
-  return utf8_to_hex_entities (s);
+
+  bool should_convert_to_hex= true; // Transcoding is required by default
+
+  // Iterate over the string and check if it contains CJK characters
+  for (int i= 0; i < N (s);) {
+    unsigned int code= decode_from_utf8 (s, i);
+
+    // Determining whether a character is a CJK character
+    if ((code >= 0x4e00 &&
```
da-liii, line 558: 这个判断在哪儿应该已经有了

src/Data/String/converter.cpp
```diff
@@ -547,7 +547,39 @@ string
 utf8_to_html (string_u8 input) {
   converter conv= load_converter ("UTF-8", "HTML");
   string    s   = apply (conv, input);
-  return utf8_to_hex_entities (s);
+
```
da-liii, line 550: 在 24_19.scm 里面，针对 utf8->html写一下单元测试

src/Data/String/converter.cpp
da-liii: 看这里： https://github.com/XmacsLabs/lolly/blob/main/lolly/data/unicode.hpp#L45

你写的可能更完整一些可以调用lolly里面的函数，在加上你自己的额外的判断

ATQlove:
Currently added 24_19.scm, which is a test for utf8->html.

Running this test before the modification results in the following:
before

Running this test after the modification results in the following:
after

ATQlove:
The file location of the call to the utf8->html function is as follows:
utf8-html

The file location of the call to the utf8_to_html function is as follows:
utf8_to_html

You can see that this modification will only affect the html export scenario and not other files.

src/Data/String/converter.cpp
ATQlove: The `is_cjk_unified_ideographs` function in `lolly/data/unicode` is used to detect strings that have been transcoded into `&#xXXXX;` form, not the original Unicode characters.

The logic of this function is as follows:

1. it iterates through the input string s, looking for character entities of the form `&#xXXXX;`.
2. Once it finds a character entity of the form `&#xXXXX;`, it extracts the XXXX part of it. 
3. It then checks to see if the XXXX is within the CJK Unified Ideographs range (4E00 to 9FBF). 
4. It returns true if all detected character entities are within this range, otherwise it returns false.
Therefore, the `is_cjk_unified_ideographs` function is not designed to be used in this place, where a judgment needs to be made before the encoding is converted.

da-liii: 接口接受的参数不一样，没啥问题了。

da-liii (approved):

PR #2066, opened 2024-09-03
[24_20] Fix the problem of missing titles and images when exporting to docx

ATQlove:
## What
Fix the problem of missing titles and images when exporting to docx.

## Why
In previous versions, exporting `tm`, `tmu` files to `docx` format would be missing titles and images. The main causes are as follows:
1. Missing title: The problem arises because string-save saves the output of `serialize-html` function, which only converts `html-stree` to `html-snippet` and is not the final output of exporting to html. Change it to `export-buffer-main` to include the export to html process completely;
2. Missing pictures: the missing pictures are mainly due to the fact that when executing the pandoc command, the execution environment is not consistent with the path of exporting to html, so the picture information can not be obtained, which has been fixed now.

## How to test your changes?
tm：
图片

html:
图片

docx:
图片


da-liii (approved):
LGTM

PR #2083, opened 2024-09-11
[70_16] Fix with node cause wrong root code and no prog root fallback

UnbSky:
## What

1. With node may cause wrong root code
2. If a prog node is not encountered, the entire document will be rendered, leading to performance issues.

## Why

1. A with node has more than one string child node, and only the last child node is the one that is actually rendered as a string node.
2. When a prog node is not encountered, it uses its own string node for rendering.

### DIFF
--- a/src/Plugins/Treesitter/lang_parser.cpp
+++ b/src/Plugins/Treesitter/lang_parser.cpp
@@ -95,7 +95,7 @@ lang_parser::get_root_node (tree t, int& start_index, int& hash_code) {
     if (!is_atomic (root)) {
       tree env_child= the_drd->get_env_child (root, tree_index, MODE, "");
       // tree prog_lan=
-      the_drd->get_env_child (root, tree_index, "prog-language", "");
+      // the_drd->get_env_child (root, tree_index, "prog-language", "");
       if (env_child == "prog") {
         root= root[tree_index];
         // cout << "Tree Index: " << tree_index << " Root: " << root << LF;
--- a/src/Plugins/Treesitter/lang_parser.cpp
+++ b/src/Plugins/Treesitter/lang_parser.cpp
@@ -132,7 +132,8 @@ lang_parser::get_data_from_root (tree root, tree line, int& start_index) {
     change_line_pos << change_index;
   }
   else {
-    for (int i= 0; i < N (root); i++) {
+    int root_N= N (root);
+    for (int i= 0; i < root_N; i++) {
       tree child_node= root[i];
       if (the_drd->is_accessible_child (root, i)) {
         // cout << "Index: " << i << " Accessible Child: " << child_node << LF;
@@ -175,7 +176,8 @@ lang_parser::get_code_from_root (tree root, string& code, string_u8& code_u8) {
     }
   }
   else {
-    for (int i= 0; i < N (root); i++) {
+    int root_N= N (root);
+    for (int i= 0; i < root_N; i++) {
       tree child_node= root[i];
       if (the_drd->is_accessible_child (root, i)) {
         // cout << "Index: " << i << " Accessible Child: " << child_node << LF;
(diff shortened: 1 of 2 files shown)


da-liii:
LGTM to me, for me: 
+ Old parser: 2931 ms
+ Tree sitter: 3056 ms

src/Plugins/Treesitter/lang_parser.cpp
```diff
@@ -103,16 +107,20 @@ lang_parser::get_root_node (tree t, int& start_index, int& hash_code) {
...
+    root= t; // Set Root To Self as Fallback
+  }
   hash_code= hash (root);
   get_data_from_root (root, t, start_index);
   return root;
 }
 
 void
 lang_parser::get_data_from_root (tree root, tree line, int& start_index) {
+  if (root->op == with_op) {
```
jingkaimori, line 121: 可以利用drd，检测某个宏的子节点是什么类型，比如with宏的子节点是变量名、变量值还是常规内容：

```cpp
root = tree(WITH, "var", "value", tree(CONCAT, "some text", "other things"));
the_drd->get_type_child(root, 0) == TYPE_VARIABLE;
the_drd->get_type_child(root, 1) == TYPE_BINDING;
the_drd->get_type_child(root, 2) == REGULAR;
```
https://github.com/XmacsLabs/moebius/blob/0f5205f27c876fe3a82b4ec3b300db9f56a543d6/moebius/drd/drd_info.hpp#L94

https://github.com/XmacsLabs/moebius/blob/0f5205f27c876fe3a82b4ec3b300db9f56a543d6/moebius/drd/tag_info.hpp#L17-L43

jingkaimori (commented):
从With之类的宏中，区分变量和文档内容的语句还能改进：

* 如果只管with，可以用预定义的`WITH`常量；
* 如果需要管其他宏，可以用drd，检测各个子节点的内容类型，区分出不需要高亮的内容

src/Plugins/Treesitter/lang_parser.cpp
UnbSky: with(color, red, main)的red类型也是TYPE_REGULAR，只有color是TYPE_VARIABLE,这个和描述好像不太一样

src/Plugins/Treesitter/lang_parser.cpp
```diff
@@ -103,16 +107,20 @@ lang_parser::get_root_node (tree t, int& start_index, int& hash_code) {
       }
     }
   }
-  // if(N(father_ip) == 1){
-  //   cout << "Can not meet Prog " << t << " IP:" << obtain_ip(t) << LF;
-  // }
+  if (N (father_ip) == 1) {
```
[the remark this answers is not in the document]
UnbSky: cpp文件目前是匹配不到prog结点的

src/Plugins/Treesitter/lang_parser.cpp
UnbSky: 使用is_accessible_child接口实现

src/Plugins/Treesitter/lang_parser.cpp
```diff
@@ -131,8 +132,12 @@ lang_parser::get_data_from_root (tree root, tree line, int& start_index) {
     change_line_pos << change_index;
   }
   else {
-    for (tree child_node : root) {
-      get_data_from_root (child_node, line, start_index);
+    for (int i= 0; i < N (root); i++) {
```
da-liii, line 135: use root_N instead of N(root)

src/Plugins/Treesitter/lang_parser.cpp
```diff
@@ -170,8 +175,12 @@ lang_parser::get_code_from_root (tree root, string& code, string_u8& code_u8) {
     }
   }
   else {
-    for (tree child_node : root) {
-      get_code_from_root (child_node, code, code_u8);
+    for (int i= 0; i < N (root); i++) {
```
da-liii, line 178: use root_N instead of N(root)

jingkaimori (approved):
LGTM

PR #2090, opened 2024-09-13
[80_3] Enhance sup/subscript kerning via OpenType math table

KeShih:
## What
We have utilized the italic correction table and kerning table in OpenType to optimize the typesetting of superscripts and subscripts.

However, it is important to note that this implementation is non-standard.

The italic correction table is used to improve the typesetting on the right side of glyphs. For general characters, the right subscript uses the default position, while the right superscript needs to be shifted to the right by this correction. For symbols like integral signs, the official documentation suggests that both the right superscript and right subscript should be moved by half the offset, while some documents recommend that the superscript remain stationary and the subscript be moved by the full offset. For aesthetic reasons, we have adopted a compromise solution here, where the subscript is shifted left by 0.6 times the offset and the superscript is shifted right by 0.4 times the offset.

For the kerning table, it provides correction values for the left and right superscripts and subscripts of a glyph based on their height. According to the OpenType standard recommendations, we should select an optimal correction value by considering both the base glyph and the superscript/subscript glyph at the two recommended correction heights. However, we are unable to handle this in the font's cpp file, so we have opted for a simpler approach, using `fn->y1` and `fn->y2` as the correction heights for the subscript and superscript, respectively, which also yields satisfactory results. 

Another fact is that many fonts do not provide a kerning table or only provide a very limited kerning table, and almost none of them provide correction values for left superscripts and subscripts. So it can be considered that this PR does not actually change any left-side typesetting.

## Why
For certain OpenType fonts, some symbols such as integral signs and superscripts/subscripts appear in incorrect positions. Now, we can improve this by utilizing the metrics provided in OpenType.

## How to test
Open `devel/80_3.tmu`.

Before:
image

After:
image

It can be seen that the first two Opentype fonts have been correctly improved, while the latter two have not changed because we used a hardcoded solution.

### DIFF
--- a/devel/80_3.tmu
+++ b/devel/80_3.tmu
@@ -1,0 +1,63 @@
+<TMU|<tuple|1.0.5|1.2.9.5>>
+
+<style|generic>
+
+<\body>
+  Asana Math:
+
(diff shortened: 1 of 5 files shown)


da-liii (approved):
approved this pull request for KeShih to move on. I'm too busy to review this pull request. FYI @mgubi 

The pull request links provided in the OSPP system must be merged and then @mgubi can approve the application.

As far as I know, these are the last two pull requests waiting to be merge:
+ https://github.com/XmacsLabs/mogan/pull/2090
+ https://github.com/XmacsLabs/mogan/pull/2120

PR #2091, opened 2024-09-14
[78_8] Fix warning: remove unused

chen-qingyu:
## What

Fix warnings

`src\Typeset\Boxes\Basic\boxes.cpp(898): warning C4551: function call missing argument list`

`src\Plugins\Tex\fromtex.cpp(2635): warning C4129: '{': unrecognized character escape sequence`

## How to test your changes?

import latex file test.tex: `123{123}\{123\}\cases{123}` shows that `"\{"` in fromtex.cpp should be `"{"` instead of `"\\{"`.

### DIFF
--- a/devel/78.tmu
+++ b/devel/78.tmu
@@ -7,7 +7,7 @@
 
   <section|Tasks>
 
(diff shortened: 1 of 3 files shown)


da-liii (approved):
Welcome

PR #2094, opened 2024-09-17
[24_22] Add documentation to remind users to use the installed version of pandoc

ATQlove:
## What
As title

## Why
To call the no-install version of pandoc must get the pandoc path in the environment variable PATH, which is not recommended.
Added a debug information to inform user which pandoc is being used.

### DIFF
--- a/TeXmacs/plugins/docx/progs/data/docx.scm
+++ b/TeXmacs/plugins/docx/progs/data/docx.scm
@@ -50,7 +50,7 @@
                                   (url->string html-temp-url)
                                   " -o "
                                   (url->string docx-temp-url))))
-          ; (display cmd) ;; For debugging
+          (display (string-append "debug: cmd for Pandoc: " cmd)) ;; For debugging
           (system cmd)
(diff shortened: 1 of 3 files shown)


TeXmacs/plugins/docx/progs/data/docx.scm
```diff
@@ -50,7 +50,7 @@
                                   (url->string html-temp-url)
                                   " -o "
                                   (url->string docx-temp-url))))
-          ; (display cmd) ;; For debugging
+          (display (string-append "debug: cmd for Pandoc: " cmd)) ;; For debugging
```
da-liii, line 53: 这里需要用 debug-io ，这样在调试控制台里面可以看到

比如：
```
TeXmacs/progs/texmacs/texmacs/tm-files.scm:                (debug-message "debug-io" (string-append msg "\n" path))
```

da-liii:
Works fine:
image

da-liii (approved):
LGTM

PR #2102, opened 2024-09-23
[70_18] Fix wrong hash code and empty first line

UnbSky:
## What

1. An incorrect `hash (tree)` can lead to an inability to recompile.
2. The first line being empty can cause rendering errors.

## Why

1. An additional item has been added to the hash_code.
2. Change `change_line_high > 0` to `change_line_high >= 0`.


jingkaimori (approved):
LGTM, bugs mentioned in lark is fixed

PR #2114, opened 2024-09-29
[80_4] Improving the Typesetting of Limit Boxes

KeShih:
## What
Improving the Typesetting of Limit Boxes.

## Why
For some OpenType fonts, the heights of the upper and lower limits in the limit box environment are not suitable.

## How to test your changes?
Open `devel/80_4.tmu`.

Before (Anasa Math font), the upper limit is too high:
image

After:
image

### DIFF
--- a/src/Graphics/Fonts/font.hpp
+++ b/src/Graphics/Fonts/font.hpp
@@ -34,6 +34,7 @@ struct font_glyphs;
 #define MATH_TYPE_NORMAL 0
 #define MATH_TYPE_STIX 1
 #define MATH_TYPE_TEX_GYRE 2
+#define MATH_TYPE_OPENTYPE 3
 
 #define START_OF_LINE 1
 #define END_OF_LINE 2
@@ -52,7 +53,7 @@ struct font_glyphs;
 
 struct font_rep : rep<font> {
   int    type;         // font type
-  int    math_type;    // For TeX Gyre math fonts and Stix
+  int    math_type;    // For TeX Gyre math fonts and Stix and OpenType fonts
   SI     size;         // requested size
   SI     design_size;  // design size in points/256
   SI     display_size; // display size in points/PIXEL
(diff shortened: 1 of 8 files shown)


src/Graphics/Fonts/font.hpp
```diff
@@ -73,6 +74,12 @@ struct font_rep : rep<font> {
   SI ysup_hi_lim;  // upper limit for supscripts
   SI yshift;       // vertical script shift inside fractions
 
+  // only for opentype fonts
```
da-liii, line 77: Moving these members to the unicode_font_rep struct?
```
struct unicode_font_rep : font_rep {
  string      family;
  int         hdpi;
  int         vdpi;
  font_metric fnm;
  font_glyphs fng;
  int         ligs;
```

src/Plugins/Freetype/tt_face.cpp
```diff
@@ -54,6 +55,17 @@ tt_face_rep::tt_face_rep (string name) : rep<tt_face> (name) {
   }
   ft_select_charmap (ft_face, ft_encoding_adobe_custom);
   bad_face= false;
+
+  // .ttf/.otf file may contain a math table
+  if (suffix (u) == "ttf" || suffix (u) == "otf") {
```
da-liii, line 60: Sometimes, the suffixes are TTF and OTF. You should use:

```
string font_suffix= locase_all (suffix (u));
if (font_suffix == "ttf" || font_suffix == "otf") { bla bla
```

src/Plugins/Freetype/tt_tools.hpp
```diff
@@ -161,7 +169,26 @@ struct MathConstantsTable {
...
+    case MathConstantRecordEnum::scriptScriptPercentScaleDown:
+      return scriptScriptPercentScaleDown;
+    case MathConstantRecordEnum::delimitedSubFormulaMinHeight:
+      return delimitedSubFormulaMinHeight;
+    case MathConstantRecordEnum::displayOperatorMinHeight:
+      return displayOperatorMinHeight;
+    case MathConstantRecordEnum::radicalDegreeBottomRaisePercent:
+      return radicalDegreeBottomRaisePercent;
+    }
+    return 0; // should never reach here
```
da-liii, line 190: you call put a `FAILED("xxx")`, and when it reaches here, there will be an exception

src/Graphics/Fonts/font.hpp
da-liii: I saw the usage in script_boxes.cpp. Because we are using struct, we can not take advance of Object Oriented Paradigm here.

da-liii:
And I tested it in your branch:
image

PR #2120, opened 2024-09-30
[80_8] Using OpenType glyph construct/assembly  table

KeShih:
# What?
Improved the typesetting of rubber character using the Opentype glyph assembly table, particularly in terms of brackets of different sizes. 
In fact, this PR have already supported all variants in the Opentype assembly table, but since Texmacs itself may not provide such variants, we cannot directly use them. We can use some unconventional methods to make it appear, such as Asana Math providing the assembly for Unicode 2907. We can use `<left-#2907-50>` to render its 50 variant (don't care why "left").
image
image

# Why?
For Opentype math fonts, many rubber characters, such as various braces, are not properly constructed.
# How to test it?
Open `devel/80_8.tmu`.

Before:
image

After:
image

It can be seen that the first two Opentype fonts have been correctly improved, while the latter two have not changed because we used a hardcoded solution.


da-liii (approved):
LGTM

PR #2129, opened 2024-10-03
[24_26] Fix for static web generation because mathml is enabled for html export

da-liii:


da-liii:
Here is the related pull request: https://github.com/XmacsLabs/html/pull/8

PR #2166, opened 2024-11-12
[59_23] Keyboard: Add std = as a keybinding for zoom-in

da-liii:
## What
Add `std =` as a keybinding for zoom-in

## Why
`C-=` for zoom in is nice than pressing `C-S-=`.

## How to test your changes?
+ [x] Linux (KDE)
+ [ ] macOS
+ [ ] Windows


da-liii:
Should be ok on all platform, will test it in RC releases.

PR #2212, opened 2025-01-04
[23_27] Fix bugs when drawing semicircle

wind1900:
## What
Fix 2 little bugs.
1. Can not draw the correct semicircle in current environment.
2. There will be "division by zero" exception in console when you're choosing the third point and move mouse to center point (the first point).
## Why
When drawing an arc, we first select the circle center `c`, then the start point `p` and end point `x` on the arc.
It will use these points to calculate a third point `m` on arc, and use `p`, `m`, `x` to draw the arc.
By default it's counterclockwise though it's called "Std-arc".
When we want to draw a semicircle, the angle between `vec-c-p` and `vec-c-q` is 180.
In the previous code, it uses `vec-c-p` as the third point `m` incorrectly.
We should still consider it's clockwise or counterclockwise here and then add the correct vector to the center point as the third point.
Before:
image

Fixed:
image

The distance of `c` and `q` as `r1` is used to avoid the exception message in console, like
```
;/: division by zero, (/ 0.0 0.0)
;    (let ((n (length opts))) (cond ((= n...
;    D:\mogan\build\packages\app.mogan\data\progs\kernel\library\base.scm, line 270, position: 0
; (let ((n (length opts))) (cond ((= n 0) (... ; f: 0.0
; (f2s (/ (point-get-x pt) f))               ; pt: (point "0.0" "0.0")
; (list-values 'point (f2s (/ (point-get-x ...
; std-arc-helper: (equal? (point-get-unit v... ; vec-c-p: (point "2.0" "0.0")
;                                              vec-c-q: (point "0.0" "0.0")
```
## How to test your changes?
Add a unit test so I can test it by `xmake run 23_27`

### DIFF
--- a/TeXmacs/progs/graphics/graphics-markup.scm
+++ b/TeXmacs/progs/graphics/graphics-markup.scm
@@ -87,26 +87,30 @@
          (p  (if (tm-point? P) (tree->stree P) c))
          (q  (if (tm-point? Q) (tree->stree Q) p))
          (r  (points-distance c p))
+         (r1 (points-distance c q))
          (x  (if (equal? r 0.0)
                 c
-                (points-add (point-times (point-get-unit (points-sub q c)) r) c)))
+                (if (equal? r1 0.0)
+                  p
+                  (points-add (point-times (point-get-unit (points-sub q c)) r) c))))
          (mid-p-x (points-mid p x))
          (vec-c-p (points-sub p c))
          (vec-c-q (points-sub q c))
-         (m  (if (equal? r 0.0)
-                c
+         (m  (if (or (equal? r 0.0) (equal? r1 0.0))
+                x
                 (if (clockwise (points-cross-product-k vec-c-p vec-c-q) 0)
                   (points-add (point-times (point-get-unit (points-sub mid-p-x c)) (- r)) c)
                   (if (= (points-cross-product-k vec-c-p vec-c-q) 0)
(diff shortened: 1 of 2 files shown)


TeXmacs/progs/graphics/graphics-markup.scm
```diff
@@ -87,28 +87,32 @@
...
-                (points-add (point-times (point-get-unit (points-sub q c)) r) c)))
+                (if (equal? r1 0.0)
+                  p
+                  (points-add (point-times (point-get-unit (points-sub q c)) r) c))))
          (mid-p-x (points-mid p x))
          (vec-c-p (points-sub p c))
          (vec-c-q (points-sub q c))
-         (m  (if (equal? r 0.0)
-                c
+         (m  (if (or (equal? r 0.0) (equal? r1 0.0))
```
da-liii, line 99: 检查一个数是否等于零，可以使用zero?这个函数

TeXmacs/progs/graphics/graphics-markup.scm
```diff
@@ -87,28 +87,32 @@
...
+                x
                 (if (clockwise (points-cross-product-k vec-c-p vec-c-q) 0)
                   (points-add (point-times (point-get-unit (points-sub mid-p-x c)) (- r)) c)
                   (if (= (points-cross-product-k vec-c-p vec-c-q) 0)
```
da-liii, line 103: 这里也可以用 zero?

da-liii (approved):

Issue #2217, opened 2025-01-09
Shift-F5 shortcut prints "S-F5" in document


woutersj:
### Mogan Version

1.2.8

### Operating System Version and Architecture

Fedora Linux 40 (GNOME Wayland)

### Describe Bug

When pressing Shift-F5 to insert a big operator like an integral sign, the characters "S-F5" are inserted in the document.

If subsequently the keyboard shortcut is completed by pressing 'I', "S-F5" is replaced by the integral sign as intended.
If anything else is pressed (that doesn't complete the shortcut sequence) the characters "S-F5" are left in the document.

In the comments on #1450 this is mentioned as being introduced in 65_8.

### Expected Behavior

No characters being inserted when Shift-F5 is pressed.

da-liii:
Fixed by https://github.com/XmacsLabs/mogan/commit/0f02cfccc793d1c8e38b527900e4d8a0d914c181

da-liii:
Author is @JimZhouZZY

PR #2233, opened 2025-06-03
[improve] Add M-a/C-a event for QTMLineEdit

JimZhouZZY:
## What  
Added support for `Cmd+A` / `Ctrl+A` key events in `QTMLineEdit` to select all text in the search toolbar.

Related: #2232

## Why  
`Cmd+A` / `Ctrl+A` is a native `QLineEdit` behavior in Qt. Supporting it improves usability by allowing users to quickly modify their search query. This standard interaction should not be ignored, I guess.

## How to test your changes  
1. Build the project.  
2. Open the search toolbar using `Cmd+f` / `Ctrl+f`.  
3. Enter any text.  
4. Press `Cmd+a` / `Ctrl+a` to ensure all text is selected.

PR #2234, opened 2025-06-03
[fix] Shift-F5 being falsely inserted as 'S-F5'

JimZhouZZY:
## What  
Resolved the issue where `Shift-F5` was incorrectly interpreted as `S-F5` by adding a simple check: `comb != "S-F5"` during fallback.

Related: #2217

## Why  
Though this may not be the most elegant solution, it effectively addresses the issue for now, I think.

## How to test your changes  
1. Build and run the project  
2. Enter math mode by typing `$`  
3. Press `Shift+F5` or `Shift+F5+I` or `Shift+F5+O` or `Shift+F5+U` or `Shift+F5+N` (possibly with the Fn key) 
4. Verify that the integral sign is inserted, rather than the string `S-F5`

### DIFF
--- a/src/Edit/Interface/edit_keyboard.cpp
+++ b/src/Edit/Interface/edit_keyboard.cpp
@@ -152,7 +152,7 @@ edit_interface_rep::try_shortcut (string comb) {
             verbatim (rhs), shorth == "" ? 1 : 3000);
     }
     if ((status & 1) == 1) cmd ();
-    else if (N (shorth) > 0) call ("kbd-insert", shorth);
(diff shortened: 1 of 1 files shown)


JackYansongLi:
这个好，我看了没问题。等明天沈达再review一下。

BTW，师弟加个微信，haggittli

JackYansongLi:
same problem as https://github.com/XmacsLabs/mogan/pull/2233

JackYansongLi:
@da-liii macOS arm64的CI挂了，貌似和我本地端一个问题，看看是不是xmake的配置有问题

da-liii (approved):

PR #2858, opened 2026-02-25
[201_85]: Prevent matrix/det from being resized by mouse drag

### LINKED ISSUE #2783: 矩阵和行列式可以通过鼠标拖动改变大小
Image

divyansharma001:
Fixes #2783

## Problem
Matrices and determinants (and other math table environments) could be resized by dragging the mouse — both via internal grid line dragging and scale handles. This behavior is incorrect since these are mathematical constructs, not regular tables.

## Changes
- **`src/Edit/Interface/edit_interface.cpp`**: Added `matrix`, `matrix*`, `bmatrix`, `Bmatrix`, `det`, `choice`, and `stack` to the `is_true_table` exclusion list, preventing scale resize handles from appearing.
- **`src/Edit/Interface/edit_mouse.cpp`**: 
  - Added `is_true_table` guard in `table_line_start` to block grid line dragging for math tables.
  - Added `is_true_table` check in `mouse_any` hover logic so the resize cursor no longer appears on math table grid lines.
- **`devel/201_85.md`**: Developer document for this task.

## How to Test
1. Insert a matrix (`Insert` -> `Table` -> `Matrix`) and click inside it
2. Verify that no resize handles appear around the matrix
3. Hover over the internal grid lines — cursor should remain normal
4. Try dragging a grid line — it should not resize
5. Repeat for `det`, `bmatrix`, `Bmatrix`, `choice`, and `stack`
6. Insert a normal table (`Insert` -> `Table` -> `Plain tabular`) — resize handles and line dragging should still work as before

### DIFF
--- a/devel/201_85.md
+++ b/devel/201_85.md
@@ -1,0 +1,21 @@
+# [201_85] Matrices and determinants should not be resizable by mouse drag
+
+## How to Test
+1. Insert a matrix (`Insert` -> `Table` -> `Matrix`) and click inside it
+2. Verify that no resize handles appear around the matrix
+3. Hover over the internal grid lines of the matrix — the cursor should remain normal (no resize cursor)
+4. Try dragging a grid line inside the matrix — it should not resize
+5. Repeat the above for `det`, `bmatrix`, `Bmatrix`, `choice`, and `stack`
+6. Insert a normal table (`Insert` -> `Table` -> `Plain tabular`) — resize handles and line dragging should still work as before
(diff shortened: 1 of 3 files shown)


JackYansongLi:
See: https://github.com/MoganLab/mogan/pull/2829

JackYansongLi:
Fixed by: https://github.com/MoganLab/mogan/pull/2829

Therefore, I close this PR.

PR #2860, opened 2026-02-25
[202_105]: Fix replace-all to replace entire document, not just after cursor

### LINKED ISSUE #2857: 搜索替换支持替换全文，而不仅仅是替换光标后的全文
目前的功能是替换所有后续匹配项

divyansharma001:
## What
Replace-all now replaces all occurrences in the entire document instead of only those after the current cursor position.

## Why
Previously, "Replace All" only replaced matches after the cursor, which is unexpected — users expect it to replace every occurrence in the document. Fixes #2857.

## How
- Call `(search-extreme-match #f)` before the replace loop in `replace-all` to navigate to the first match
- Updated tooltip strings from "Replace all further occurrences" to "Replace all occurrences"
- Updated Chinese comment in search-kbd.scm accordingly

## How to test
1. Open a document with multiple occurrences of a word (e.g. "hello" appearing 5 times)
2. Place cursor in the middle of the document (after the 3rd occurrence)
3. Open search and replace (Edit → Search and replace)
4. Enter search term "hello" and replacement "world"
5. Click "Replace all" (Ctrl+Enter / Command+Enter)
6. Confirm all 5 occurrences are replaced, not just the 2 after the cursor


wumoin:
“Replace All Occurrences” should be implemented as a separate button, while keeping the original “Replace all further occurrences” behavior.

wumoin:
Thanks for the contribution!
The related files have changed in the past few days.
Could you please update this PR based on the latest main branch and adjust the implementation accordingly?

divyansharma001:
- devel/202_105.md now includes both the prior "Meta" platform-key fix (merged Feb 26) and the new "Replace all occurrences" button entry with test steps.
- Commit messages cleaned up (fixed typo in "occurrences").
- No further conflicts after resolution.

Implementation keeps original behavior as requested. Ready for review/merge — let me know if any adjustments needed (e.g., tooltip translations or buffer handling tweaks from Copilot feedback).
Thanks!

devel/202_105.md
```diff
@@ -31,3 +31,42 @@ else
 ```
 
 Also removed the now-unused `("meta::keyboard" "")` entry from `TeXmacs/plugins/lang/dic/en_US/zh_CN.scm`.
+
+# [202_105] Search and replace: add "Replace all occurrences" button
```
wumoin: The documentation should be moved to a new file instead of being placed in `devel/202_105.md`.

divyansharma001:
Done, moved the "Replace all occurrences" documentation to a new file devel/202_110.md and removed it from devel/202_105.md. Let me know if you'd prefer a different task number for the new file.

PR #2865, opened 2026-02-25
[206_19]: Fix Latin Modern Math cannot input "h" in math mode

### LINKED ISSUE #2742: 数学字体 Latin Modern Math 无法输入"h"
如果设置数学模式的字体为 Latin Modern Math 则无法在数学模式下输入"h"

divyansharma001:
Fixes #2742

## Problem
When the math mode font is set to Latin Modern Math, typing "h" in math mode produces no output.

## Root Cause
In `smart_font_rep::advance()`, lowercase letters are mapped to Unicode Mathematical Italic codepoints via a simple offset: `0x1D44E + (s[0] - 'a')`. For 'h', this computes U+1D455 — but U+1D455 is intentionally **unassigned** in the Unicode standard. The correct math italic small h is at **U+210E** (PLANCK CONSTANT) in the Letterlike Symbols block. Since Latin Modern Math (and all conforming OpenType Math fonts) has no glyph at U+1D455, the character fails to render.

## Fix
- Add a special case in `smart_font_rep::advance()` to map 'h' → U+210E instead of the unassigned U+1D455
- Add the corresponding reverse mapping `U+210E → 'h'` in `init_unicode_substitution()` for consistency

## How to Test
1. Set the math mode font to Latin Modern Math
2. Enter math mode and type "h"
3. Confirm "h" displays correctly as math italic h

Developer document: `devel/206_19.md`

### DIFF
--- a/devel/206_19.md
+++ b/devel/206_19.md
@@ -1,0 +1,22 @@
+# 206_19 修复 Latin Modern Math 字体在数学模式下无法输入 "h" 的问题
+
+## 如何测试
+1. 将数学模式字体设置为 Latin Modern Math
+2. 进入数学模式，输入字母 "h"
+3. 确认 "h" 能正常显示为数学斜体 h（U+210E）
+
+## 2026/02/26 修复数学斜体 h 的 Unicode 映射
+### What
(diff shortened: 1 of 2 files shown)


da-liii:
Here is the behavior after the fix:
image

da-liii (approved):
LGTM

Issue #2886, opened 2026-02-27
在 right footer 可以显示符号对应的快捷键


da-liii:

Prthmsh7:
> No description provided.

Can you please assign this issue to me, I would love to work on this one

Prthmsh7:
Hi @da-liii , I’ve pushed the clang-format fix and updated the PR. I’d really appreciate your feedback when you get a chance. Thanks!

Prthmsh7:
This issue can be closed now.

PR #2892, opened 2026-02-27
[202_106] display the shortcut keys corresponding to the symbols in the right footer

Prthmsh7:
PR for #2886

**Before :**
Screenshot 2026-02-28 at 2 17 45 AM

**After :**
Screenshot 2026-02-28 at 2 32 56 AM

**Root cause:** The footer had no mechanism to look up or display keyboard shortcuts 
**Fix:** Added `get_shortcut_suffix()` which calls Scheme's kbd-find-inv-binding to resolve the shortcut for symbols, and used the existing kbd_shortcut() for operations which appends the result as [Shortcut] in the right footer.

## How to Test
1. Open a document in the editor.  
2. Move the cursor to a position where a symbol is present.  
3. Verify that the right footer displays the corresponding keyboard shortcut.  
4. Move the cursor away from the symbol and confirm that the footer clears.

### DIFF
--- a/src/Edit/Interface/edit_footer.cpp
+++ b/src/Edit/Interface/edit_footer.cpp
@@ -348,10 +348,6 @@ edit_interface_rep::compute_operation_footer (tree st) {
     }
   }
   if (last_item (tp) == 0) r= concat ("before ", r);
-  string op_name   = drd->get_name (L (st));
-  tree   shortcut  = kbd_shortcut ("(make '" * op_name * ")");
-  string shortcut_s= as_string (shortcut);
-  if (shortcut_s != "") r= concat (r, " [", shortcut, "]");
   return r;
 }
 


MoonL79:
The modifications in `compute_operation_footer()` don't seem to serve any purpose. I tried deleting this part and the functionality still works fine. Do you understand what your modifications in `compute_operation_footer` actually mean?

Prthmsh7:
@MoonL79 You’re right. I initially tried to extend the shortcut display to structural operations by resolving their insertion command (e.g.. (make 'frac)). However, this turned out to be unnecessary for the symbols requirement in #2886, since the shortcut display is already handled in compute_text_footer. I’ve removed the redundant lines and updated the PR.  Thanks for pointing it out, please let me know if you have any further feedback.

MoonL79:
Please resolve the numbering conflict, and then I will merge your changes.

JackYansongLi:
please also change the PR number

Issue #2894, opened 2026-02-28
在 macOS 平台保存文件的对话框中无法使用 Command v 粘贴


da-liii:
临时解决方案，先随便挑一个名字，再重命名

git-lakshy:
This is a known Qt bug with native macOS file dialogs (NSOpenPanel/NSSavePanel). When Qt uses the native macOS dialog, keyboard shortcuts for editing operations don't work properly in the filename text field.

another approach could be to use Qt's custom dialog box on macos, I tried and it works but with slight ui tradeoff

![Image](https://github.com/user-attachments/assets/11c3db55-96e4-410b-b3ce-640eb0bb32b8)

PR #2901, opened 2026-03-01
[202_106]: fix Cmd+V paste not working in macOS save file dialog

divyansharma001:
## What
Fix standard editing shortcuts (Cmd+V, Cmd+C, Cmd+X, Cmd+A) not working in
the native file dialog on macOS.

## Why
On macOS, QAction shortcuts registered in the menu bar are converted to
NSMenuItem key equivalents. When a native file dialog (NSSavePanel/NSOpenPanel)
is shown, these key equivalents intercept standard editing shortcuts before
the dialog's text field can handle them — so pressing Cmd+V triggers Mogan's
"Paste" action instead of pasting into the filename field.

## How
Temporarily disable all QAction shortcuts before `dialog->exec()` and restore
them after the dialog closes. This allows the native dialog to process
Cmd+V/C/X/A through the standard Cocoa responder chain.

### How to test
1. Open Mogan on macOS
2. Copy some text to the clipboard
3. File → Save as
4. Press Cmd+V in the filename field — text should be pasted
5. Verify Cmd+C, Cmd+X, Cmd+A also work in the dialog
6. After closing the dialog, verify editor shortcuts still work normally

Fixes: https://github.com/XmacsLabs/mogan/issues/2894


Yuki-Nagori:
Change the number, it‘s conflicting with another PR

Yuki-Nagori:
I tested it, and I still can't paste when saving the file.

divyansharma001:
Sorry for that, will look into this again.

divyansharma001:
The previous approach (clearing QAction shortcuts) didn't work because Qt converts shortcuts to NSMenuItem key equivalents in the native Cocoa menu bar, and clearing the Qt-side shortcuts doesn't remove those already-cached native key equivalents.

New approach: Temporarily remove the entire native menu via [NSApp setMainMenu:nil] before dialog->exec(), then restore it after. This removes all NSMenuItem key equivalents from the Cocoa responder chain, allowing the native file dialog to process Cmd+V/C/X/A normally.

PR #2915, opened 2026-03-03
[205_2009] Fix inconsistent document rendering when closing tabs

### LINKED ISSUE #2009: Inconsistent document rendering
### Mogan 版本

1.2.8

### 操作系统版本和架构

Microsoft Windows Version 22H2 (OS Build 19045.2965)

### 描述问题

In the case of multiple tabs, the rendering effect of the same document is inconsistent in different windows.

**Steps to reproduce**
- Open Mogan
- Set `Edit` > `Preferences...` > `Buffer management` > `Documents in separate windows`
- Set `View` > `Tab bar`
- Type `集合与映射`
- Set Font to `Microsoft YaHei`, and size to 14pt
- Create a new document
- Close the new document in Tabs

**Preview**
image

### 期待的结果

Consistent rendering effect

Utkal059:
## Problem
Fixes #2009

When using "Documents in separate windows" mode with the Tab bar enabled, 
closing a newly created tab causes the original document's rendering to 
become inconsistent — the same text appears with different font metrics 
and spacing than before.

## Root Cause
In `kill_window()` inside `src/Texmacs/Data/new_window.cpp`, when the 
last tab in a window is closed, `set_current_view()` switches to the 
surviving window's view. But `set_current_view()` only updates the view 
pointer — it never calls `resume()` on the editor. 

Without `resume()`, `notify_change(THE_FOCUS + THE_EXTENTS)` is never 
triggered, so the font environment and page layout are not recalculated. 
This leaves the editor in a stale state, producing the visual mismatch.

There was already a FIXME comment at this location noting that GUI focus 
was not properly transferred.

## Fix
Added `resume()` and `send_keyboard_focus()` after `set_current_view()` 
in `kill_window()`, matching the pattern already used by 
`var_focus_on_buffer()` in `new_view.cpp`.

## Developer Doc
See `devel/205_2009.md`

### DIFF
--- a/devel/205_2009.md
+++ b/devel/205_2009.md
@@ -1,0 +1,23 @@
+# 205_2009 Fix inconsistent document rendering when closing tabs
+
+## Problem
+Issue #2009: When using "Documents in separate windows" mode with Tab bar enabled, closing a tab causes the remaining document's rendering to become inconsistent (different font rendering for the same content).
+
+## Root Cause
+In `kill_window()` (`src/Texmacs/Data/new_window.cpp`), when the last tab in a window is closed, `set_current_view()` is called to switch to the surviving window's view. However, `set_current_view()` only updates the view pointer — it does not call `resume()` on the editor. Without `resume()`, `notify_change(THE_FOCUS + THE_EXTENTS)` is never triggered, so the font environment and layout are not recalculated, resulting in inconsistent rendering.
+
+## Fix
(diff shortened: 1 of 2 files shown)


MoonL79:
Have you reproduced this issue on your device? I am unable to reproduce it on mine.
This issue is quite old, and many changes have occurred in the current version. If you can reproduce the problem, please provide a video.

MoonL79:
Sorry, it's my fault. This issue is too old and seems to have become invalid, so I have closed it.

Issue #2910, opened 2026-03-03
数学模式下波浪号、方括号、竖线没有Tab循环框


Yuki-Nagori:
<img width="29" height="58" alt="Image" src="https://github.com/user-attachments/assets/923e6804-3315-44ce-825c-23da37853686" />

JerryMainBlog:
i'll see that

1sh-repalto:
Hello, I am able to reproduce this behavior. I'll start working on this.

Prashant27-07:
Hi, I'd like to work on this issue. I'm a GSOC 2026 applicant familiar with C++. Can I take this?

Prthmsh7:
@Prashant27-07 I think @1sh-repalto is already working on this one.

1sh-repalto:
Hello @Prashant27-07, I am working on this. I will raise the PR soon, meanwhile you can take up some other issue.

Prashant27-07:
@1sh-repalto ok

PR #2939, opened 2026-03-05
[201_90] fix inner product symbol display in tab cycle popup

### LINKED ISSUE #2926: Inner product symbol does not show on Tab cycling popup window
Inner product symbol does not show on Tab cycling popup window as shown in the image below:

Image

1sh-repalto:
Fixes #2926 
Inner product symbol does not show on Tab cycling popup window

## Summary
When cycling through math variants for the `<` key (which includes the inner product `⟨⟩` as the 7th variant), the popup window showed a broken/red symbol instead of the angle brackets.

**Developer Document:** `devel/201_90.md`

## Issue Found
The `math-bracket-open` case in `lambda-to-symbol` was concatenating the left and right bracket symbols (e.g., `"<langle><rangle>"`). This produced a compound string that the `widget-box` rendering engine (using the `roman` font) could not render correctly, showing as a red broken symbol or `--`.

## Changes
In `TeXmacs/progs/math/math-edit.scm`:

Modified `lambda-to-symbol` to return only the left bracket symbol for `math-bracket-open`:

```diff
 ((math-bracket-open)
  (and (>= (length (cdr body)) 2)
       (string? (cadr body))
       (string? (caddr body))
       (let ((lb (cadr body))
             (rb (caddr body)))
         `(symbol-completion
-          ,(string-append lb rb)))))
+          ,lb))))
```

### How to test
1. Open Mogan Editor
2. Enter math mode by pressing `$`
3. Type `<`
4. Press `Tab` repeatedly (7 times) to reach the inner product `⟨⟩` variant
5. Verify that the popup window shows the left angle bracket `⟨` clearly (instead of a broken red symbol or `--`)
6. Press `Tab` once more to return to `<` and verify the cycle continues correctly

Screenshot 2026-03-05 145152

### DIFF
--- a/TeXmacs/progs/kernel/gui/menu-widget.scm
+++ b/TeXmacs/progs/kernel/gui/menu-widget.scm
@@ -825,6 +825,24 @@
             ((== p '()) p)
             (else (list (make-menu-bad-format p style))))))
 
+(define (make-symbol-completion-widget sym col)
+  ;; Renders a math symbol using the full TeXmacs typesetting engine
+  (let* ((math-content `(math ,sym))
+         (doc `(document 
+                 (with "color" ,col 
+                       "font-size" "5"   
+                       "page-type" "user"
+                       "page-odd" "3px"     
+                       "page-even" "3px"  
+                       "page-top" "3px"          
+                       "page-bot" "3px"            
+                       ,math-content)))
+         (style `(style (tuple "generic")))
+         (output-wid (widget-texmacs-output doc style)))
+    (widget-menu-button output-wid
+                        (make-menu-command (insert sym))
+                        "" "" 0)))
+
 (define-table make-menu-items-table
   (glue (:boolean? :boolean? :integer? :integer?)
         ,(lambda (p style bar?)
@@ -846,12 +864,12 @@
                (list (make-menu-symbol p style '() (color symbol-color))))))
   (symbol-completion (:string? :*)
           ,(lambda (p style bar?)
-             (let ((symbol-color (if (== (get-preference "gui theme") "liii-night") "white" "black")))
-               (list (make-menu-symbol p style '(roman mr medium normal 10 600 0) (color symbol-color))))))
+              (let ((symbol-color (if (== (get-preference "gui theme") "liii-night") "white" "black")))
+                (list (make-symbol-completion-widget (cadr p) symbol-color)))))
   (symbol-completion* (:string? :*)
           ,(lambda (p style bar?)
-             (let ((symbol-color (if (== (get-preference "gui theme") "liii-night") "#ff6666" "red")))
-               (list (make-menu-symbol p style '(roman mr medium normal 10 600 -2) (color symbol-color))))))
+              (let ((symbol-color (if (== (get-preference "gui theme") "liii-night") "#ff6666" "red")))
+                (list (make-symbol-completion-widget (cadr p) symbol-color)))))
   (texmacs-output (:%2)
     ,(lambda (p style bar?) (list (make-texmacs-output p style))))
   (texmacs-input (:%3)
--- a/src/Texmacs/Window/tm_button.cpp
+++ b/src/Texmacs/Window/tm_button.cpp
@@ -296,16 +296,16 @@ texmacs_output_widget (tree doc, tree style) {
   // SI dw2= env->get_length (PAGE_SCREEN_RIGHT);
   // SI dh1= env->get_length (PAGE_SCREEN_BOT);
   // SI dh2= env->get_length (PAGE_SCREEN_TOP);
-  color col= env->get_color (BG_COLOR);
-  if (env->get_string (BG_COLOR) == "white" &&
-      is_transparent (extract (doc, "body")))
+  color col  = env->get_color (BG_COLOR);
+  bool  trans= is_transparent (extract (doc, "body"));
+  if (env->get_string (BG_COLOR) == "white" && trans)
 #ifdef QTTEXMACS
     col= rgb_color (236, 236, 236);
 #else
     col= light_grey;
 #endif
   double zoom= (retina_zoom == 2 ? 1.0 : 1.2);
-  return widget (tm_new<box_widget_rep> (b, col, false, zoom, 0, 0));
+  return widget (tm_new<box_widget_rep> (b, col, trans, zoom, 0, 0));
 }
 
 array<SI>


JackYansongLi:
Could you make it show $\langle \rangle$ instead of a single $\langle$

wumoin:
image
The Tab cycling for the `[` symbol seems to have some issues.

1sh-repalto:
Hello @JackYansongLi and @wumoin , 
Thank you for your reviews. I noticed that these issues had already been fixed when I worked on #2910. I created the PR #2916 which fixed these tab cycling behaviors and it was merged. However, a later PR #2918 modified `font_fn` in `src/Texmacs/Window/tm_button.cpp`:

```diff
- font fn= smart_font (family, fn_class, series, shape, sz, dpi);
+ font fn= find_font (family, fn_class, series, shape, sz, dpi);
```

From my investigation, this change led to the unusual behaviour of the fixes I had done in my PR.
**From my perspective, the best and minimal change would be to revert `find_font` back to `smart_font`**. I also attempted to resolve the issue while keeping `find_font`, but I was unable to find a clean solution. Is there some good reason to let it stay as `find_font` and also possible to fix the issues at the same time ?

Let me know what you guys think on this.

I have tested both behaviours and am attaching them for reference:

1. with `smart_font`

https://github.com/user-attachments/assets/48f32a64-11e7-467a-a905-b212bf56e7eb

2. with `find_font`

https://github.com/user-attachments/assets/172dd7fd-cd56-482b-a8aa-52040d559b55

wumoin:
> Hello @JackYansongLi and @wumoin , Thank you for your reviews. I noticed that these issues had already been fixed when I worked on #2910. I created the PR #2916 which fixed these tab cycling behaviors and it was merged. However, a later PR #2918 modified `font_fn` in `src/Texmacs/Window/tm_button.cpp`:
> 
> ```diff
> - font fn= smart_font (family, fn_class, series, shape, sz, dpi);
> + font fn= find_font (family, fn_class, series, shape, sz, dpi);
> ```
> 
> From my investigation, this change led to the unusual behaviour of the fixes I had done in my PR. **From my perspective, the best and minimal change would be to revert `find_font` back to `smart_font`**. I also attempted to resolve the issue while keeping `find_font`, but I was unable to find a clean solution. Is there some good reason to let it stay as `find_font` and also possible to fix the issues at the same time ?
> 
> Let me know what you guys think on this.
> 
> I have tested both behaviours and am attaching them for reference:
> 
> 1. with `smart_font`
> 
>  with_smart_font.mp4 
> 2. with `find_font`
> 
>  with_find_font.mp4




Using smart_font causes the program to crash and also leads to incorrect rendering.
The following shows the correct display.
image
After switching to `smart_font`, the display becomes inaccurate.
image
Additionally, clicking this option will also cause the program to crash.
image

1sh-repalto:
Thanks for the clarification @wumoin. I'll look into this issue again and try solve it with `find_font`.

1sh-repalto:
Hello @JackYansongLi @wumoin, I have modified the code to fix the rendering of tab cycle components. Here is a quick summary of what I did:

Instead of forcing the basic C++ `box_widget` to parse complex macros, this fix changes the Scheme logic to use `widget-texmacs-output`. This safely routes the popup symbols through the editor's full math typesetting engine, ensuring complex macros like `<langle>` render perfectly. I also fixed a hardcoded opacity bug in `tm_button.cpp` so the tab cycle symbols have a transparent background.
 
Let me know if this approach looks good. If so, I will update the developer document and PR description accordingly.

**Note**: While fixing this issue I found some symbols that don't render properly in tab cycling but it's not a new issue as I checked on old version too they don't work properly. I'll raise new issues to fix them.
 
 Here is the state of tab cycling after this fix:

https://github.com/user-attachments/assets/49474255-5cde-4666-b5ab-e466e274feb6

wumoin:
The current changes significantly degrade the performance of tab cycling. In particular, the response becomes noticeably slower when pressing Tab to cycle through the options.

You can observe the difference in the before-and-after video comparison. When holding down the Tab key continuously, the interface becomes noticeably laggy.

https://github.com/user-attachments/assets/611b1ab3-537d-4abc-938a-ab485919d765

1sh-repalto:
Hi @wumoin, thank you for pointing this out. My bad I didn't catch this.
I have addressed this performance lag by introducing caching inside `texmacs_output_widget` in `tm_button.cpp`.
The lag was caused by `widget-texmacs-output` re-initializing a full typesetting environment for every symbol on every Tab press. The new implementation uses a static hashmap keyed by the `(doc, style)` tree pair to cache the computed boxes. This ensures each unique symbol (e.g., <, {}, etc.) is typeset exactly once; all subsequent cycles retrieve the pre-rendered box instantly from memory.

Does this seem like a correct approach ?

Below is the code diff. I have also attached a screen recording demonstrating the before vs. after performance. 

Inside `tm_button.cpp`
```diff
+// Cached result for texmacs_output_widget
+struct texmacs_output_cache_entry {
+  box   b;
+  color col;
+  bool  trans;
+};
+
 widget
 texmacs_output_widget (tree doc, tree style) {
+  // --- Cache lookup ---
+  static hashmap<tree, texmacs_output_cache_entry> output_cache;
+  tree cache_key= tuple (doc, style);
+  if (output_cache->contains (cache_key)) {
+    texmacs_output_cache_entry& e= output_cache (cache_key);
+    double zoom= (retina_zoom == 2 ? 1.0 : 1.2);
+    return widget (tm_new<box_widget_rep> (e.b, e.col, e.trans, zoom, 0, 0));
+  }
+
   .
   .
   .
+
+  // --- Cache store ---
+  texmacs_output_cache_entry entry;
+  entry.b    = b;
+  entry.col  = col;
+  entry.trans= trans;
+  output_cache (cache_key)= entry;
+
   double zoom= (retina_zoom == 2 ? 1.0 : 1.2);
   return widget (tm_new<box_widget_rep> (b, col, trans, zoom, 0, 0));
 }
```

https://github.com/user-attachments/assets/79e73c5a-1c3c-417f-994f-41942a500440

wumoin:
Initializing each symbol through a full typesetting environment is not a very good approach.

I think this issue should be split into two separate parts: one is the display problem with `[` and `|`, and the other is the issue where `<` cannot be displayed in tab cycling.

I just took a look at the first problem, and I will submit a PR for it later. You can take a look at that then.

wumoin:
#2973

wumoin:
You can look into how these symbols are rendered. I’ll close this PR for now.

1sh-repalto:
Got it, I'll create a new PR that addresses this issue based on the changes you have made in your PR. Should I wait for your PR to get merged first ?

wumoin:
#2569 #2574 #2589 
I gave you a few references. You can first look into how these symbols are rendered, and then investigate why one of the tab-cycling entries for < is not being rendered. This is somewhat difficult, so you may want to try working on other issues first.

PR #2953, opened 2026-03-07
[201_95] Disable command mode autocomplete suggestions in macro editor

### LINKED ISSUE #2873: 宏编辑器中应该禁用命令模式 \ 的补全选项
可以通过 (current-buffer) 的  url 来判断

1sh-repalto:
Fixes #2873 
In the macro editor, the autocomplete suggestions in command mode `⟨\⟩` should be disabled.

## Summary
Disabled the autocomplete popup suggestions in the macro editor while keeping the `\` hybrid command mode functional.

## Issue Found
When editing macros in the macro editor, typing `\` would trigger the command mode  `⟨\⟩` with a popup of autocomplete suggestions.

## Changes
Modified `src/Edit/Interface/edit_source.cpp`:

```cpp
// Disable autocomplete options in the macro editor
string buf_name= as_string (buf->buf->name);
if ((starts (buf_name, "tmfs://aux/edit-") &&
     buf_name != "tmfs://aux/edit-shortcuts" &&
     !starts (buf_name, "tmfs://aux/edit-comment")) ||
    buf_name == "tmfs://aux/macro-editor") {
  return;
}
```

### How to test
1. Open Mogan Editor.
2. Open the macro editor via **Tools → Macros → New macro or Edit macros**.
3. Press the `\` key and begin typing.
4. **Verify**:
   - The `⟨\⟩` bracketed tag (hybrid command mode) appears correctly.
   - *No* autocomplete suggestion popup appears while typing.
5. Test in a normal document:
   - **Verify**: The `\` key triggers the hybrid command mode completely, and autocomplete functions as normal.
6. Test in search/replace window:
   - **Verify**: The `\` key triggers the hybrid command mode completely, and autocomplete functions as normal.

https://github.com/user-attachments/assets/5ff49c4c-5734-40b6-b7a1-b3cea03d0507

### DIFF
--- a/TeXmacs/progs/generic/generic-kbd.scm
+++ b/TeXmacs/progs/generic/generic-kbd.scm
@@ -25,12 +25,6 @@
         (texmacs texmacs tm-print)
         (doc help-funcs)))
 
-(tm-define (inside-macro-editor?)
-  (with u (url->string (current-buffer))
-  (or (== u "tmfs://aux/macro-editor")
-    (and (string-starts? u "tmfs://aux/edit-")
-      (not (== u "tmfs://aux/edit-shortcuts"))))))
-
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; General shortcuts for all modes
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -48,7 +42,7 @@
   ("]" (make-bracket-close "]" "["))
   ("{" (make-bracket-open "{" "}"))
   ("}" (make-bracket-close "}" "{"))
-  ("\\" (if (or (inside? 'hybrid) (in-prog?) (inside-macro-editor?)) (insert "\\") (make-hybrid)))
+  ("\\" (if (or (inside? 'hybrid) (in-prog?)) (insert "\\") (make-hybrid)))
   ("\\ var" "\\")
   ("\\ var var" "<setminus>")
   ("$" (make 'math))
--- a/src/Edit/Interface/edit_source.cpp
+++ b/src/Edit/Interface/edit_source.cpp
@@ -18,6 +18,7 @@
 #include "merge_sort.hpp"
 #include "observers.hpp"
 #include "preferences.hpp"
+#include "tm_buffer.hpp"
 #include "tree_observer.hpp"
 
 #include <moebius/data/scheme.hpp>
@@ -30,6 +31,15 @@ void
 edit_interface_rep::source_complete_try () {
   bool is_source     = (get_env_string ("mode") == "src");
   bool is_source_mode= (get_env_string ("preamble") == "true");
+
+  // Disable autocomplete options in the macro editor
+  string buf_name= as_string (buf->buf->name);
+  if ((starts (buf_name, "tmfs://aux/edit-") &&
+       buf_name != "tmfs://aux/edit-shortcuts" &&
+       !starts (buf_name, "tmfs://aux/edit-comment")) ||
+      buf_name == "tmfs://aux/macro-editor") {
+    return;
+  }
 
   if (is_source && (!is_source_mode)) {
     completion_style= get_preference ("completion style");
--- a/devel/201_95.md
+++ b/devel/201_95.md
@@ -1,40 +1,35 @@
-# [201_95] Disable hybrid command mode in macro editor
+# [201_95] Disable autocomplete options in macro editor
 
 ### What
-Disabled the hybrid command mode (the `⟨\⟩` tag) in the macro editor. This prevents the `\` key from triggering command mode and its associated autocomplete suggestions.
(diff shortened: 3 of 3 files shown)


wumoin:
What is needed is to disable the entire `\` command behavior.

src/Edit/Interface/edit_source.cpp
```diff
@@ -30,6 +31,16 @@ void
 edit_interface_rep::source_complete_try () {
   bool is_source     = (get_env_string ("mode") == "src");
   bool is_source_mode= (get_env_string ("preamble") == "true");
+
+  // Disable autocomplete in the macro editor
+  string buf_name= as_string (buf->buf->name);
+  if ((starts (buf_name, "tmfs://aux/edit-") &&
```
wumoin: This condition would also prevent `\` from working in other windows, such as the `comment-editor`. I checked `TeXmacs/progs/source/macro-widgets.scm`, 和 it seems that the corresponding tmfs path is not complete. As a result, the path never contains `macro-editor`. You could try modifying it by directly changing it to `tmfs://aux/edit-macro-editor`, which would allow it to target the `macro-editor `specifically.

1sh-repalto:
Thank you for the clarification @wumoin! I had asked about this in comments as well as in discord but didn't receive any response so just went with this approach. I'll update the PR asap.

1sh-repalto:
Hello @wumoin, I have refactored the logic and also updated the developer document as well as the PR body. Please review when you have time.

TeXmacs/progs/generic/generic-kbd.scm
```diff
@@ -42,7 +42,7 @@
   ("]" (make-bracket-close "]" "["))
   ("{" (make-bracket-open "{" "}"))
   ("}" (make-bracket-close "}" "{"))
-  ("\\" (if (or (inside? 'hybrid) (in-prog?)) (insert "\\") (make-hybrid)))
+  ("\\" (if (or (inside? 'hybrid) (in-prog?) (has-style-package? "macro-editor")) (insert "\\") (make-hybrid)))
```
wumoin: You cannot use `(has-style-package? "macro-editor")` for this check, because it would also disable the `\` command in other windows, such as search/replace.

You can use `(current-buffer)` to determine whether the current context is inside the macro editor. For reference, you can look at the implementation of `(inside-search-buffer?)` in `TeXmacs/progs/generic/search-widgets.scm`.

1sh-repalto: Thanks for the feedback! 

I have updated the implementation to use `current-buffer` instead. I added an `inside-macro-editor?` predicate in `generic-kbd.scm` that checks if the buffer URL matches `tmfs://aux/macro-editor or tmfs://aux/edit-*` (explicitly excluding `tmfs://aux/edit-shortcuts`).
The \ command now correctly inserts a literal backslash in the macro editor while keeping the hybrid command mode fully functional everywhere else, including the search and replace input buffers.

I've pushed the updated changes! Let me know if further changes are required.

TeXmacs/progs/generic/generic-kbd.scm
```diff
@@ -25,6 +25,12 @@
         (texmacs texmacs tm-print)
         (doc help-funcs)))
 
+(tm-define (inside-macro-editor?)
+  (with u (url->string (current-buffer))
+  (or (== u "tmfs://aux/macro-editor")
+    (and (string-starts? u "tmfs://aux/edit-")
+      (not (== u "tmfs://aux/edit-shortcuts"))))))
```
wumoin:
```
(tm-define (inside-macro-editor?)
  (with u (url->string (current-buffer))
  (or (== u "tmfs://aux/macro-editor")
    (and (string-starts? u "tmfs://aux/edit-")
      (not (== u "tmfs://aux/edit-shortcuts"))
      (not (string-starts? u "tmfs://aux/edit-comment"))))))
```
The check for `edit-comment` is missing.

Also, sorry for the confusion. The original author of the issue just told me that what actually needs to be disabled is the autocomplete options, not the entire \ command behavior. Sorry for the inconvenience and for taking up your time.

1sh-repalto: > Sorry for the inconvenience and for taking up your time.
 

No worries at all! In fact, thank you for taking the time to review the code and provide such helpful guidance.

Just for a final clarification: the aim is to disable the autocomplete popup suggestions **only** in the macro editor, but keep them fully enabled in the normal editor, search, replace, and other auxiliary editors (like comments and shortcuts). The `\` key should also continue to behave normally everywhere.

Does that sound correct?

wumoin: > > Sorry for the inconvenience and for taking up your time.
> 
> No worries at all! In fact, thank you for taking the time to review the code and provide such helpful guidance.
> 
> Just for a final clarification: the aim is to disable the autocomplete popup suggestions **only** in the macro editor, but keep them fully enabled in the normal editor, search, replace, and other auxiliary editors (like comments and shortcuts). The `\` key should also continue to behave normally everywhere.
> 
> Does that sound correct?

Yes, that understanding is correct.

1sh-repalto: > Yes, that understanding is correct.

Cool! I have made the necessary changes as per the final clarifications, updated the developer document `devel/201_95.md` and the PR body. I have also attached a screen recording of the fix in the PR body. 
Please review when you have time and let me know if further changes are required.

PR #2957, opened 2026-03-07
[201_94]: Fix slink bare URL redirect without HTTPS (#2324)

### LINKED ISSUE #2324: \slink 不用https，无法正确跳转 liiistem.cn
输入\slink,点击Enter键，输入网址liiistem.cn，点击无法跳转，需要加https

Image

按住ctrl+点击操作
Image

想要的效果：不用加https就可以点击跳转

divyansharma001:
## Problem
When typing `slink` and entering a bare domain like `liiistem.cn` (without `https://`), Ctrl+clicking the link does nothing. The URL is treated as a local file path because `go-to-url` routes it through `default-root-handler`, which fails to find any local file.

## Fix
Added detection of bare domain-like URLs in `go-to-url` (`TeXmacs/progs/link/link-navigate.scm`). When the input string:
- Has no protocol prefix (`http://`, `https://`, etc.)
- Is not a file path (doesn't start with `/`, `.`, `~`, `#`)
- Contains a dot (`.`)
- Does not exist as a local file

...then `https://` is automatically prepended before URL routing.

## How to test
1. Type `slink`, press Enter, enter `liiistem.cn`
2. Ctrl+click the rendered link
3. Browser should open `https://liiistem.cn`

Also verify these still work correctly:
- `https://liiistem.cn` — unchanged, opens normally
- `www.google.com` — opens as `https://www.google.com`
- `./local-file.tm` — not affected (relative path)

Closes #2324

### DIFF
--- a/TeXmacs/progs/link/link-navigate.scm
+++ b/TeXmacs/progs/link/link-navigate.scm
@@ -497,12 +497,27 @@
   ; TODO: jump to anchors in HTML
   (noop))
 
+(define (url-looks-like-bare-domain? u)
+  "Check if a default-rooted URL looks like a bare domain (e.g. liiistem.cn)"
+  (let* ((s (url->system u)))
+    (and (not (string-null? s))
+         (not (string-starts? s "/"))
+         (not (string-starts? s "."))
+         (not (string-starts? s "~"))
+         (not (string-index s #\space))
+         (not (string-index s #\/))
+         (string-index s #\.)
+         (not (url-exists? (url-expand u))))))
+
 (define (default-root-handler u)
   (cond ((and (url-rooted-protocol? u "default")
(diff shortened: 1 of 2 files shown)


wumoin:
There are still many issues. For example, if the link points to a file, it may be incorrectly identified as a web URL.

divyansharma001:
Fixed. Moved the bare-domain detection from go-to-url into default-root-handler, so it only triggers after normal file resolution has been attempted.

The previous approach checked file existence with system->url (relative to working directory), which could misidentify file links like notes.pdf as web URLs. Now:

-> Relative path resolution (url-relative (current-buffer)) runs first
-> url-expand path resolution runs next
-> Only if neither finds a local file, and the string looks like a bare domain (has a dot, no slashes), does it fall back to https://

PR #2958, opened 2026-03-07
[203_26]: fix LaTeX export images in inaccessible formats

divyansharma001:
- Add magic header detection for images with unknown suffixes
- Add jpeg to fast-path suffix list in tmtex-as-eps
- Copy files with correct extension when format is LaTeX-compatible
- Add fallback to print-snippet when convert-to-file fails
- Register postscript-file converters for jpeg, gif, tif via ImageMagick


AXeonV:
image

I insert a jpeg and export the document as LaTeX. Then it throws error above.


Firstly, I think the key point for the LaTeX exporting bug is that, when inserting a embedded image and converting the document into LaTeX version, the raw data of the image will be base64-coded so that the data flow will be transformed between C++ and Scheme.

I've digged into the bug and find that the image data had been base64-coded **twice** in the memory somwhere. When converting, it will call

```cpp
url
get_from_ramdisc (url u) {
  if (!is_ramdisc (u)) return url_none ();
  url res= get_cache (u);
  if (!is_none (res)) return (res);
  url tmp= url_temp (suffix (u));
  save_string (tmp, u[1][2]->t->label);
  return set_cache (u, tmp);
}
```

and save the coded image data into system temp path. But I locate that file and find that it had been base64-coded **twice**. So when converting this image to pdf version, it crashed.

It seems that your fix haven't fix the bad data in the temp path when LaTeX converting.

Secondly, I think jpg and png formats are more common than jpeg, tif and gif. You could fix jpg and png embedded image when LaTeX converting first.

Lastly, thank you for this PR. This bug is very critical.

divyansharma001:
Thank you for the detailed analysis! You were right, the root cause was double base64 encoding.

I've pushed two additional fixes:

Root cause: object_l1.cpp::tmscm_to_content was not decoding base64 when receiving (raw-data ...) from Scheme, so the C++ RAW_DATA tree ended up storing base64 text instead of binary. This base64 string then got written to the temp file via get_from_ramdisc, causing Qt/MuPDF to fail reading it.

Fixes:

moebius/Scheme/L1/object_l1.cpp: Added base64 decode in tmscm_to_content for RAW_DATA nodes, so the Scheme→C++ boundary always produces binary in RAW_DATA
TeXmacs/progs/generic/embedded-edit.scm: Encode binary image data as base64 before storing into (raw-data ...) when embedding an image, to match the decode above
This establishes a clear invariant: Scheme layer always holds base64 in raw-data, C++ layer always holds binary in RAW_DATA.

Regarding your second point, I've also kept jpg and png in the fast path from the previous commit. 

Happy to prioritize those formats further if needed.

JackYansongLi:
please resolve the conflicts.

AXeonV:
I recommend splitting this PR into two separate PRs.

One PR would address the fix for the double Base64 encoding issue during LaTeX conversion. I think it only involves changes to two files: `object_l1.cpp` and `embedded-edit.scm`.

The other PR would cover the magic header detection and the other features currently included in this PR.

This separation would make the review and testing process clearer and more manageable.

PR #2960, opened 2026-03-08
Align TEXMACS_HOME_PATH Fallbacks With XDG/Platform Defaults (fixes #771 path inconsistency)

### LINKED ISSUE #771: Plugin Center
Users should open a gui from mogan, which lists all plugin available for mogan. Then users select plugin to install, update or modify its setting.

Design of this tool should be discussed:
1. Where is plugins host by community
We should build a service to host plugin archives
2. Where is plugins stored on users' device
Install to their user directory responding to each os, such as `%APPDATA%` under windows
3. How to fetch and install plugins
**(Need Discussion)** We could fetch via [Qt Installer Framework](https://doc.qt.io/qtinstallerframework/)
4. Is it possible to hot-update plugins?
**(Need Discussion)** Depends on module system
5. should we provide R7RS and SRFI interface for plugin?
**(Need Discussion)**

Vatsalpatni73:
### Summary
This PR fixes an inconsistency in home-path fallback behavior that was still relevant to issue #771 (plugin/user data location policy).

Some startup and cache code paths still defaulted to `~/.TeXmacs`, while newer logic uses platform-specific/XDG-compliant locations. This caused mismatched behavior for plugin and user-data directories, especially on Linux.

---

### What changed
- Updated startup default path selection in `init_texmacs.cpp`.
- Updated `get_texmacs_home_path()` fallback in `tm_sys_utils.cpp`.
- Updated cache initialization fallback in `data_cache.cpp`.

---

### New fallback behavior

- **Windows:** `$APPDATA/<PREFIX_DIR>`
- **macOS:** `$HOME/Library/Application Support/<PREFIX_DIR>`
- **Linux/Unix:** `$XDG_DATA_HOME/<PREFIX_DIR>`  
  *(fallback: `$HOME/.local/share/<PREFIX_DIR>`)*

- **wasm:** `./<PREFIX_DIR>`

---

### Why this matters

- Keeps plugin and user-data paths consistent across startup/runtime/cache code.
- Removes legacy `~/.TeXmacs` fallback behavior that conflicts with XDG-based policy.
- Reduces cross-platform path surprises and aligns with the direction discussed in #771.

---

### Validation

- File diagnostics for modified files: **no errors**.
- `src/**` no longer contains `~/.TeXmacs` / `$HOME/.TeXmacs` fallbacks.

### DIFF
--- a/src/System/Boot/init_texmacs.cpp
+++ b/src/System/Boot/init_texmacs.cpp
@@ -300,14 +300,22 @@ init_main_paths () {
   url default_path;
 #if defined(OS_MINGW) || defined(OS_WIN)
   default_path= get_env ("APPDATA") * "/" * PREFIX_DIR;
+#elif defined(OS_MACOS)
+  default_path= get_env ("HOME") * "/Library/Application Support/" *
+                PREFIX_DIR;
+#elif defined(OS_WASM)
+  default_path= string ("/.") * PREFIX_DIR;
 #else
-  default_path= "~/.TeXmacs";
+  string xdg_data_home= get_env ("XDG_DATA_HOME");
+  if (is_empty (xdg_data_home)) xdg_data_home= get_env ("HOME") * "/.local/share";
+  default_path= xdg_data_home * "/" * PREFIX_DIR;
 #endif
   url home_path= get_env_path ("TEXMACS_HOME_PATH", default_path);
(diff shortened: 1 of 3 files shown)


wumoin:
What is the goal of this change? The current caching mechanism does not appear to have any issues.

PR #2974, opened 2026-03-12
[222_54] fix: override tooltip to Toggle cell alignment in tabular context only.

### LINKED ISSUE #2852: The "numbered" toggle button for tabular inside tables is misleading — it controls alignment, not numbering
### Description: 
When the cursor is inside a tabular within a big-table or small-table environment, the focus toolbar shows a "numbered/unnumbered" toggle button (tm_numbered.xpm / 123-style icon) with the tooltip "Toggle numbering". However, this button does not control the table's numbering.

- What users expect: Clicking the "123" button should toggle the table caption between numbered ("表格 1." / "Table 1.") and unnumbered ("表格." / "Table.").
- What actually happens: The button switches between tabular and tabular*, which toggles center alignment of cells, not numbering.
- How numbering actually works: Table numbering is controlled at the outer level by toggling between big-table and big-table* (or small-table / small-table*), which already works correctly.
 
From TeXmacs/packages/standard/std-markup.ts (line 320):

```scheme
<assign|tabular*|<macro|body|<tformat|<cwith|1|-1|1|-1|cell-halign|c>|<arg|body>>>>
```

This confirms that tabular* only sets cell-halign to center — it has nothing to do with numbering.

## Proposed Fix

Change the icon and/or tooltip of the toggle button in the tabular context so it accurately reflects its actual purpose (cell alignment toggle), rather than appearing as a numbering control.

## Context

Discovered during investigation in PR #2833 / issue #2787. As clarified by @Yuki-Nagori, the original issue was based on a misunderstanding of what the button does. The underlying functionality works correctly — only the UI representation is misleading.

## Related

Issue: #2787
PR: #2833

Prashant27-07:
Fixes #2852

The tooltip "Toggle numbering" was misleading - the button actually toggles cell alignment (tabular vs tabular*). Override alternate-first-icon, alternate-second-icon, alternate-first-name and alternate-second-name in table-menu.scm to show correct icons and tooltip only in tabular context using table-markup-context?.

### DIFF
--- a/TeXmacs/progs/table/table-menu.scm
+++ b/TeXmacs/progs/table/table-menu.scm
@@ -668,3 +668,19 @@
 	((balloon (icon "tm_cell_join.xpm")
 		  "Join selected cells")
 	 (cell-set-span-selection)))))
+
+(tm-define (alternate-first-icon t)
+  (:require (table-markup-context? t))
+  "tm_cell_left.xpm")
+
+(tm-define (alternate-second-icon t)
+  (:require (table-markup-context? t))
+  "tm_cell_center.xpm")
+
+(tm-define (alternate-first-name t)
+  (:require (table-markup-context? t))
+  "Toggle cell alignment")
+
+(tm-define (alternate-second-name t)
+  (:require (table-markup-context? t))
+  "Toggle cell alignment")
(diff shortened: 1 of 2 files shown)


Prashant27-07:
Hey @JackYansongLi  @Yuki-Nagori I raised another PR for same issue hope this will has no conflict at all because i checked all the thing this time.

PR #2981, opened 2026-03-15
[201_96] fix: accent marks tooltip shows option key ⌥ on macOS

### LINKED ISSUE #2962: macOS 上重音符号快捷键显示有问题
Image

divyansharma001:
## Summary

Fixes #2962

- On macOS, the keyboard shortcut tooltip for accent marks (e.g. the tilde accent `~`) was showing a wrong symbol (⊙) instead of the Option key symbol ⌥
- Root cause: `kbd_system_prevails` in `tm_config.cpp` was converting `"A-~"` (Option+~) to `"escape ~"` for display, causing the Escape key symbol to render as ⊙ in the apple-lucida font at small size
- Fix: simplified `kbd_system_prevails` to return the input unchanged, so `A-` is decoded as `⌥` (U+2325) via the existing macOS keyboard map

## Test plan

- [ ] Open Mogan on macOS
- [ ] Enter math mode and open the accent insert popup
- [ ] Hover over the tilde accent button
- [ ] Verify tooltip shows `等效快捷键: (⌥ ~)` with the correct Option key symbol ⌥ instead of ⊙

### DIFF
--- a/devel/201_96.md
+++ b/devel/201_96.md
@@ -1,0 +1,28 @@
+# [201_96] fix: accent marks tooltip shows option key ⌥ on macOS
+
+## How to Test
+
+1. Open Mogan on macOS.
+2. Enter math mode and open the accent insert popup (click the `~` accent button in the toolbar).
+3. Hover over any accent button (e.g., the tilde accent `tm_tilda.xpm`).
+4. The tooltip should display `等效快捷键: (⌥ ~)` — the Option key symbol ⌥ followed by `~`.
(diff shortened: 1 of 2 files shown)


JackYansongLi:
CI failed, please fix it

Yuki-Nagori:
Thank you, I looked at your code. Making this change might affect Windows and Linux. I think https://github.com/MoganLab/mogan/pull/2982 is better, so this PR is closed. Sorry.

PR #2982, opened 2026-03-15
[222_60] Correct macOS modifier symbols in accent shortcut tooltips

Prthmsh7:
fixes: #2962 
This patch ensures that modifier symbols are correctly rendered in Qt tooltips on macOS through two small adjustments:

### 1. Tooltip UTF-8 Conversion
In `src/Plugins/Qt/qt_ui_element.cpp`, tooltip text is now passed through `cork_to_utf8()` before being converted to a Qt string. This converts TeXmacs character references into their proper Unicode symbols so that Qt displays them correctly.

### 2. macOS Shortcut Rendering Condition
In `src/Texmacs/Server/tm_config.cpp`, the `kbd_system_prevails` condition was refined to avoid applying legacy modifier substitutions when macOS fonts are already being used. This prevents conflicts with the font-based rendering of macOS modifier symbols.

After:
Screenshot 2026-03-16 at 2 04 06 AM


Prthmsh7:
@MoonL79 please review when free, Thanks!

Prthmsh7:
@JackYansongLi  please review when free, thanks!

Yuki-Nagori:
The number needs to be updated. I checked the code and there's no problem.

Prthmsh7:
@Yuki-Nagori Updated numbering in a follow-up commit. Please review when free. Thanks!

PR #3104, opened 2026-04-04
[201_104] Fix inner product symbol not rendering on Tab cycling popup window

### LINKED ISSUE #2926: Inner product symbol does not show on Tab cycling popup window

1sh-repalto:
Fixes #2926 
Inner product symbol does not show on Tab cycling popup window

## Summary
When cycling through math variants for the `<` key, the inner product `⟨⟩` showed as a broken red symbol in the Tab cycling popup instead of rendering correctly.                                                                 
                                                                            
**Developer Document:** `devel/201_104.md`
                                                                           
## Root Cause
`math_font_rep`'s `get_extents`, `get_xpositions`, and `draw_fixed` all passed the full string to `search_font`, which is designed for single math symbols only. For a bracket pair like `"<langle><rangle>"` (two concatenated TeXmacs symbols), `search_font` fail — it is not a single dictionary entry and the character fallback rejects `<` — so it returns `error_font`, which  draws in red.                                                              

## Changes                                                                  
In `src/Graphics/Fonts/math_font.cpp`:
                                                                              
Added a single-char fast path (via `tm_char_forwards`) to `get_extents`, `get_xpositions`, and both `draw_fixed` overloads. Single characters use the original `search_font` lookup unchanged. Multi-character strings now iterate character by character:
                                                                             
- `get_extents` — accumulates per-symbol metrics with correct advance widths and ink bounds
- `get_xpositions` — fills position array by accumulating per-symbol advance widths         
- `draw_fixed(x, y)` — draws each symbol at its accumulated x-offset       
- `draw_fixed(x, y, xk)` — delegates to `font_rep::draw_fixed` (base class already iterates char-by-char)
                                                                              
No Scheme-level changes needed — `lambda-to-symbol` in `math-edit.scm` already returns `(string-append lb rb)`, which is the correct representation.                                                             
                  
## How to test                                                             
1. Open Mogan Editor
2. Enter math mode by pressing `$`                                          
3. Type `<`
4. Press `Tab` repeatedly to cycle through variants                         
5. **Verify**: The `⟨⟩` entry displays correctly (not in red) in the popup
6. Select `⟨⟩` and **verify** the correct angle bracket pair is inserted in the document

## Before

image

## After

https://github.com/user-attachments/assets/261ae336-0dee-4f0e-8eb5-a0ce2a543a83

### DIFF
--- a/devel/201_104.md
+++ b/devel/201_104.md
@@ -1,0 +1,27 @@
+# [201_104] Fix inner product symbol not rendering in Tab cycling popup
+
+### What
+Fixed `⟨⟩` (angle bracket pair) showing as a red broken symbol in the Tab cycling completion popup when typing `<` in math mode.
+
+### Why
+The Tab cycling popup renders symbol labels using `math_font_rep`. Its `get_extents`, `get_xpositions`, and `draw_fixed` methods all passed the full string to `search_font`, which only handles single math symbols. For a bracket pair like `"<langle><rangle>"` (two concatenated TeXmacs symbols), `search_font` fails — it is not a single dictionary entry and the character fallback rejects `<` — so it returns `error_font`, which draws in red.
+
+### How
+Modified `src/Graphics/Fonts/math_font.cpp`.
+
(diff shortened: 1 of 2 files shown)


MoonL79 (approved):
LGTM

PR #3156, opened 2026-04-16
[212_6] 修复pdf导出功能在遇到不可见字符后出现的问题

da-liii:


src/Plugins/Pdf/pdf_hummus_renderer.cpp
```diff
@@ -1331,11 +1331,24 @@ requires_hack_notdef_for_tex_font (string fontname) {
   return r;
 }
 
+static bool
+pdf_is_invisible_internal_char (int ch) {
```
da-liii: codepoint

da-liii: return codepoint == 0x200B || codepoint == 0x2060;

wumoin (approved):

PR #3256, opened 2026-04-29
[201_104] Fix : Shift+Tab results in empty popup window on `>`#2726

### LINKED ISSUE #2726: tab cycling上还有一些符号没有被显示出来
Image
如图，进入数学模式输入`<`，随后按下shift tab，最后一个符号没有在上面显示出来

Image
`>`对应的shift tab是空的

Dylan-end:
Fixes #2726
Shift+Tab results in empty popup window on `>`

## Summary
When typing `>` in math mode and pressing `Shift+Tab` to cycle backwards through variants, the Tab cycling popup window appears empty instead of showing the expected symbol variants.
Developer Document: devel/201_104.md

The backwards cycling (Shift+Tab) for the `>` key fails to display any variant in the popup.

## Root Cause
The keyboard event handling for `Shift+Tab` in math mode does not correctly trigger the variant cycling mechanism for the `>` symbol. Specifically:

1.Shift+Tab is treated as an independent key combination and is not correctly mapped to the "previous variant" operation in the math editing context.
2.When '>' is the first symbol in the variant menu and the user presses Shift+Tab, the system's handling method is to convert it into multiple 'Tab' presses and iterate through < + n 'Tab's to check for existence, until it reaches the last symbol in the variant menu.
3.In the initial definition of < + 6 Tab presses, the last entry consists of two characters and requires special handling. However, such handling is currently only implemented for math-bracket-open. Therefore, < + 6 Tab presses should be changed to math-bracket-open.

As a result, the popup window is empty, and the user cannot see any selectable options.

## How to test
1. Open Mogan Editor.
2. Enter math mode by pressing `$`.
3. Type `>` (greater-than sign).
4. **Expected behavior before fix:** Pressing `Shift+Tab` shows an empty popup window (or no popup at all).
5. Press `Shift+Tab` repeatedly to cycle backwards through variants.
6. **Verify:** The popup window displays the correct variants in order.

## Before (Problem)
image

- When pressing `Shift+Tab` on `>`, the popup window is completely empty.

## After (Fix)
image

- `Shift+Tab` now shows the variant list correctly.

## Additional Notes
- Also fixed similar >> issues.


Dylan-end:
Update: The uploaded files have already been formatted.

MoonL79 (approved):
LGTM

PR #3425, opened 2026-05-20
[0409] 添加 \varomega 命令支持

yinyuscloor:


da-liii:
see https://github.com/MoganLab/mogan/pull/3418

PR #3707, opened 2026-06-10
[0650] fix html export logic for nested decorated blocks inside framed blocks

JackYansongLi:
Fixes an HTML export bug where nested decorated blocks inside framed blocks fail to export or lose style attributes.


MoonL79 (approved):
LGTM

PR #3847, opened 2026-06-28
[0655] fix html export mathml mathvariant mapping for mathscr/cal* family

JackYansongLi:
Fixes a bug where calligraphic and script mathematical symbols in the mathscr/cal* family are exported as '?' in MathML.


yinyuscloor (approved):
LGTM

PR #4452, opened 2026-08-30
[4448] fix: allow style switching back to Beamer without saving.

adi-coderr:
### Summary
Fixes a bug where switching a document's style from Beamer to another style (e.g., Source), and then immediately back to Beamer without saving, resulted in an "operation not allowed" block.

### Changes Made
* Modified `TeXmacs/progs/generic/document-style.scm` to safely handle the Beamer style transition state.
* Updated `TeXmacs/progs/generic/test/generic-test.scm` to include an automated regression test for this specific workflow.

### Verification
* [x] Manually verified in the Mogan GUI that the style switches smoothly without errors.
* [x] Automated Scheme tests pass locally.

https://github.com/user-attachments/assets/50e0a4f5-501a-48db-b8cc-1acdc82dc3a6

### DIFF
--- a/TeXmacs/progs/generic/document-style.scm
+++ b/TeXmacs/progs/generic/document-style.scm
@@ -115,6 +115,13 @@
   ) ;when
 ) ;tm-define
 
+(tm-define (beamer-style-switch-allowed?)
+  (or (document-empty?)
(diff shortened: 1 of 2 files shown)


adi-coderr:
fixes ##4448

